PR #14, opened 2017-08-02
[Attempt to Fix Unable to Add Tags For Newly Created User]

riandyrn:
Hello, Gene

I found 2 bugs on `filterTags()` implementation which located on `session.go`:
- `dst` param should have type `*[]string` not `[]string`, the reason is because you intend to populate `dst` with filtered tags, so the param should pass by reference, not by value.
- the value should be appended to `dst` is `s` not `parts[1]`, so for example if `s` is `"email:someone@example.com"` then the value should be appended to `dst` is `"email:someone@example.com"` not `"someone@example.com"`.

The first bug will prevent new users to add tags to its account since the resulted filtered tags would always be empty.

The consequence of second bug is obvious I think.

Let me know your opinion.

Thanks.

### DIFF
--- a/server/session.go
+++ b/server/session.go
@@ -536,7 +536,7 @@ func (s *Session) acc(msg *ClientComMessage) {
 
 		if msg.Acc.Tags != nil && len(msg.Acc.Tags) > 0 {
 			tags := make([]string, 0, len(msg.Acc.Tags))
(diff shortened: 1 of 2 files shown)


or-else:
Thanks, good catch!

PR #17, opened 2017-08-04
[Attempt to Fix Incorrect Acc P2P Permission's Default Value]

riandyrn:
Hello, Gene

I found the bug. When creating the account, both `user.Access.Auth` & `user.Access.Anon` should have default access for `types.TopicCat_P2P` not `types.TopicCat_Me`.

The reason is because the `user.Access.Auth` & `user.Access.Anon` in this context is used for defining default access of current user to `P2P` topic not to `me` topic ([see here](https://github.com/tinode/chat/blob/master/API.md#acc)).

Let me know your opinion.

Thanks.

### DIFF
--- a/server/session.go
+++ b/server/session.go
@@ -515,8 +515,8 @@ func (s *Session) acc(msg *ClientComMessage) {
 		var user types.User
 		var private interface{}
 		if msg.Acc.Desc != nil {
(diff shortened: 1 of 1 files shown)


or-else:
I'll merge but there is another bug there. The assignment of default access values should be above `if msg.Acc.Desc != nil {` otherwise it's never assigned when Desc is nil.

PR #18, opened 2017-08-08
[Attempt to Fix Missing Value in {pres} Packet P2P]

riandyrn:
Hello, Gene

I found bug on {pres} packet sent by server when current user is currently not subscribing to P2P topic but user in other side send message (use case 6): the `pres.src` value is always empty for P2P topic.

The cause of this bug is because in P2P topic, the `t.x_original` value is set to "" [(see here)](https://github.com/tinode/chat/blob/devel/server/hub.go#L543), but there is no assignment value when there is a message from user on the other side. Thus this situation lead into empty value of `pres.src`.

Let me know you opinion.

Thanks.

### DIFF
--- a/server/pres.go
+++ b/server/pres.go
@@ -175,8 +175,8 @@ func (t *Topic) presAnnounceToUser(uid types.Uid, what string, seq int, list []i
 }
 
 // Announce to all/offline only subscribers on 'me' topic
-func (t *Topic) presAnnounceToSubscribers(what string, seq int, offlineOnly bool) {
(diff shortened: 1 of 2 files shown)


or-else:
Great catch, thanks! It's indeed a bug. But I would rather fix it in a simpler way. I'll send you a pull request for review.

or-else:
Here https://github.com/tinode/chat/compare/presence-fix?expand=1
Please take a look.

or-else:
Oops, sorry. It's too big. Give me a sec, I'll fix it.

or-else:
Here: https://github.com/tinode/chat/pull/19

PR #19, opened 2017-08-08
fix for the src being empty in some presence notifications

or-else:


riandyrn:
You're awesome, Gene! 👍 👍 
I'm just aware that the function of `t.original` is for generating appropriate topic name:sweat_smile:.

PR #21, opened 2017-08-13
[Attempt to Ease Migration]

riandyrn:
Hello, Gene

I would like to suggest adding safeguard for `‘A’` permission when user subscribing to `P2P` topic.

This will make our user which already installed previous version of Tinode able to migrate easily to current version without needing to upgrade their existing database which maybe already contains many users or `P2P` topics data. 

The difference between current permission model & previous one (from the data point of view) is only addition of `‘A’` permission in `P2P` topics. Yet this single permission could make such big difference in recent version. Thus in this attempt, I made sure that when user subscribing to `P2P` topic, the resulted mode would always contain `‘A’` permission unless the mode is `’N’`. If it doesn’t contain `‘A’` permission, then it means it comes from the old database, thus the server will update it.

Let me know your opinion.

Thanks

### DIFF
--- a/server/hub.go
+++ b/server/hub.go
@@ -573,7 +573,7 @@ func topicInit(sreg *sessionJoin, h *Hub) {
 			sreg.created = !user1only
 		}
         
-        // make sure modeWant & modeGiven always have 'A' permission on each subscribers
+        // make sure modeWant & modeGiven always have 'A' permission for both subscribers whether in existing or new P2P topic
         for uid, userData := range t.perUser {
             shouldUpdateSubscription := false
             modes := []*types.AccessMode{&userData.modeWant, &userData.modeGiven}
@@ -593,7 +593,11 @@ func topicInit(sreg *sessionJoin, h *Hub) {
                     "ModeGiven": int(userData.modeGiven),
                     "ModeWant": int(userData.modeWant),
                 }
-                store.Subs.Update(t.name, uid, update)
+                if err = store.Subs.Update(t.name, uid, update); err != nil {
+                    log.Println("hub: databse error in updating user subscription '" + t.name + "' (" + err.Error() + ")")
+                    sreg.sess.queueOut(ErrUnknown(sreg.pkt.Id, t.x_original, timestamp))
+                    return
+                }
                 
                 // modify current mode
                 t.perUser[uid] = userData
--- a/server/hub.go
+++ b/server/hub.go
@@ -524,6 +524,18 @@ func topicInit(sreg *sessionJoin, h *Hub) {
 					users[u2].Access.Auth,
 					types.ModeCP2P)
 			}
+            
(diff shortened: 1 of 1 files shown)


or-else:
This is a one-time change probably need by your organisation only. Why not just update your DB with a small separate program?

riandyrn:
To be honest this is also attempt to close backdoor in Tinode. Notice that in current version there is no checking to `'A'` permission for resulting mode of `P2P` topic. Thus it is still possible for creating "system-like" account by just setting the defacs without `'A'` permission, not by using Tinode application itself, but rather by tampering the db (for example by the bribed db administrator). If we add the checking, we could prevent such scenario which resulted in more robust application.

So I just thought maybe it will be useful for another Tinode user too if they want to use Tinode on production.

or-else:
But if someone gains unauthorized access to the DB they can do whatever they want regardless of permissions. This is not a backdoor, more of an issue of compromised organisation. Technical measure cannot replace adequate organizational security.

or-else:
BTW, I don't understand this comment:

> Notice that in current version there is no checking to 'A' permission for resulting mode of P2P topic.

Could you please elaborate. Thanks!

riandyrn:
> Could you please elaborate. Thanks!

If we have user A with defacs `"JRWP"` & user B with defacs `"JRWPA"`, then they create P2P connection, the resulted mode for user A would be `"JRWPA"` & the resulted mode for user B would be `"JRWP"` because there is currently no integrity check for resulted mode in P2P connection. Basically it's just that.

riandyrn:
At least if you don't feel it is necessary to perform integrity check for resulted mode in P2P connection + the use case won't be happening in the general case (maybe it's only specific to us), I think it's better if we add a section in [instruction page](https://github.com/tinode/chat/blob/master/INSTALL.md) for old Tinode users to tell them explicitly that they need to update their database. Otherwise, they might be confused.

or-else:
> If we have user A with defacs "JRWP" & user B with defacs "JRWPA"

But `JRWP` is not supposed to happen. Just use RethinkDB's Dataexplorer and update all incorrect values at once https://www.rethinkdb.com/docs/administration-tools/ 
If you need help with RethinkDB query language, please let me know, I'll help you write the queries.

BTW, I think I understand where some confusion is coming from. I intentionally made it possible to set just the `N` permission in addition to anything with an `A` in it, like `JRWPA`. `N` means "none", no access. It's a way to restrict users spamming with invitations to connect. If the option of setting the N permission is removed then it would open a path to for user harassment with announcements.  "

riandyrn:
Hehe.., never mind this case, bro 👍
Let's just move to something else, we still have a lot of things to patch 😆 👍 👍 
Btw thank you so much for replying my comments, it really make me feel appreciated 😄 👍 👍

PR #34, opened 2017-10-29
[Attempt to Fix Uninitialized recvId & readId in P2P Topic Creation]

riandyrn:
Hello, Gene

I found that on `P2P` topic creation `recvId` & `readId` attributes on `t.perUser` is not properly initialized. This will cause `readId` to become 0 when user initially only update `recvId` using `{note}`.

Let me know your opinion.

Thanks


or-else:
Thanks!

or-else:
Also here: https://github.com/tinode/chat/commit/56fae533bfa5a596bec60e657dd8ad5039211e86

PR #39, opened 2017-11-09
Plugins

or-else:


or-else:
@riandyrn Please take a look at this functionality. It allows for:
 * Creation for chatbots https://github.com/tinode/chat/tree/plugins/chatbot - one is currently running at https://api.tinode.co/x/example-react-js/
 * Command-line access to the server https://github.com/tinode/chat/tree/plugins/tn-cli
 * Option to create Tinode clients in a lot of common languages, like C++ or Objective C.

PR #40, opened 2017-11-22
[Attempt to Fix `hub.topics` Race Condition]

riandyrn:
Hello, Gene

When I hit Tinode with high traffic, I got following errors: 

- `fatal error: concurrent map read and map write`
- `fatal error: concurrent map writes`

The cause of these errors is because there is a race condition on `hub.topics`. So to fix this issue I suggest we use `sync.RWMutex` to synchronize read & write operations to the map. 

Let me know your opinion.

Thanks

### DIFF
--- a/server/hub.go
+++ b/server/hub.go
@@ -12,6 +12,7 @@ import (
 	"expvar"
 	"log"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/tinode/chat/server/store"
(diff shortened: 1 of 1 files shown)


or-else:
Thanks for the catch! This is fine, I'll merge it. But I will probably rewrite it to use https://golang.org/pkg/sync/#Map

BTW, I've been on vacation for the past week, I'll be back tomorrow.

or-else:
https://github.com/tinode/chat/commit/ff968ab711569206137f49ecd9475bb95a534ea3

PR #41, opened 2017-11-22
[Attempt to Fix Stall Shutdown]

riandyrn:
Hello, Gene

I found problem when shutting down system while it being hit by high traffic. It's stall, unable to finish the cleanup for topics. 

The issue is triggered by these lines: https://github.com/tinode/chat/blob/devel/server/hub.go#L207-L213. 

Notice that when we issue shutdown, apparently we didn't stop the hub for putting more topics to `hub.topics` (because we use goroutine to run `topicInit()`). So the result of `len(h.topics)` on [line 207](https://github.com/tinode/chat/blob/devel/server/hub.go#L207) could be different from the result of `len(h.topics)` on [line 211](https://github.com/tinode/chat/blob/devel/server/hub.go#L211). 

It is often that the result of `line 211` is greater than the result of `line 207` which then caused stall because `<-topicsdone` is always blocking.

So to overcome this issue, I suggest we prevent put operation on `hub.topics` when system shutdown is in progress.

Let me know your opinion.

Thanks

### DIFF
--- a/server/hub.go
+++ b/server/hub.go
@@ -78,6 +78,9 @@ type Hub struct {
 	// Request to shutdown, unbuffered
 	shutdown chan chan<- bool
 
+	// Flag for marking system shutdown is in progress
+	isShutdownInProgress bool
+
 	// Exported counter of live topics
 	topicsLive *expvar.Int
 }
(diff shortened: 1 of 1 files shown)


or-else:
Thanks!

or-else:
https://github.com/tinode/chat/commit/ef3c9d3c6ec8bb8eac191921a745e2e8fc65836d

PR #43, opened 2017-11-24
run hub.topicInit in hub's main go routine

or-else:
Run `hub.topicInit` in the main hub's go routine eliminating the need for locking `hub.topics`

### DIFF
--- a/server/hub.go
+++ b/server/hub.go
@@ -12,7 +12,6 @@ import (
 	"expvar"
 	"log"
 	"strings"
-	"sync"
 	"time"
 
 	"github.com/tinode/chat/server/store"
@@ -59,7 +58,7 @@ type metaReq struct {
 type Hub struct {
 
 	// Topics must be indexed by name
-	topics *sync.Map
+	topics map[string]*Topic //*sync.Map
 
 	// Channel for routing messages between topics, buffered at 4096
 	route chan *ServerComMessage
@@ -79,31 +78,25 @@ type Hub struct {
 	// Request to shutdown, unbuffered
 	shutdown chan chan<- bool
 
-	// Flag for indicating that system shutdown is in progress
-	isShutdownInProgress bool
-
 	// Exported counter of live topics
 	topicsLive *expvar.Int
 }
 
 func (h *Hub) topicGet(name string) *Topic {
-	if t, ok := h.topics.Load(name); ok {
-		return t.(*Topic)
-	}
-	return nil
+	return h.topics[name]
 }
 
 func (h *Hub) topicPut(name string, t *Topic) {
-	h.topics.Store(name, t)
+	h.topics[name] = t
 }
 
 func (h *Hub) topicDel(name string) {
-	h.topics.Delete(name)
+	delete(h.topics, name)
 }
 
 func newHub() *Hub {
 	var h = &Hub{
-		topics: &sync.Map{}, //make(map[string]*Topic),
+		topics: make(map[string]*Topic),
 		// this needs to be buffered - hub generates invites and adds them to this queue
 		route:      make(chan *ServerComMessage, 4096),
 		join:       make(chan *sessionJoin),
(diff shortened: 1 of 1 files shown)


or-else:
@riandyrn Please take a look. This seems to be a cleaner solution to the race conditions during sutdown.

riandyrn:
Hello, Gene

Yeah, it seems cleaner. But from performance perspective, I think it’s not good idea though.

The reason is because if we run `topicInit()` in hub’s main, it will be blocking until `topicInit()` finished, right? 

While the performance change seems negligible when the system only having small number of newly added connections, but on high numbers, our system performance would suffers. So I think it’s best if we keep `topicInit()` in separate goroutine.

This is what I thought. What do you think?

or-else:
Can you test it? You probably have some infrastructure for load testing.

Ultimately, `topicInit` still has to use the same number of CPU cycles regardless of being run in one or multiple go routines. Unless the bottleneck can be reduced by running in parallel (for instance, the bottleneck is in the IO *and* it can be effectively parallelized - I'm not sure it's the case), the performance should be better in case of a single go routine.

What's needed here is a load test to be sure which approach is better.

riandyrn:
> for instance, the bottleneck is in the IO and it can be effectively parallelized - I'm not sure it's the case

Yeah, in fact that’s the very case I thought.

Notice that in `topicInit()` we require several access to database before we are able to put topic in `hub.topics`:
- https://github.com/tinode/chat/blob/devel/server/hub.go#L288
- https://github.com/tinode/chat/blob/devel/server/hub.go#L307
- https://github.com/tinode/chat/blob/devel/server/hub.go#L387
- https://github.com/tinode/chat/blob/devel/server/hub.go#L453
- https://github.com/tinode/chat/blob/devel/server/hub.go#L576
- https://github.com/tinode/chat/blob/devel/server/hub.go#L597
- https://github.com/tinode/chat/blob/devel/server/hub.go#L597
- https://github.com/tinode/chat/blob/devel/server/hub.go#L719

Yet in every database access there must be latency, right? Especially if we use centralized database like DynamoDB for example.

Yet we only run single instance of `hub` loop which will receive & process all the incoming signal to our system. So in case there are a lot of incoming request for joining new topic, then the system will block until `topicInit()` is done before processing the new request, right?

But if we use goroutine, we could just process new request without waiting call to `topicInit()` finished.

> Can you test it? You probably have some infrastructure for load testing.

Sure, I’ll let you once I’m finished. Maybe I’ll do the test on Monday.

or-else:
> Sure, I’ll let you once I’m finished. Maybe I’ll do the test on Monday.

Thanks!

riandyrn:
Hello, Gene

Here is the result of `topicInit` benchmarking for each method:
- on same goroutine as hub: `48.20 sec / session` on average, `1mn 38sec / session` at highest
- on different goroutine: `0.21 sec / session` on average, `0.21 sec / session` at highest

Here is the scenario I was using to perform the test:
- login to 10,000 unique users with arrival rate of 200 new users / second
- for each connected user, subscribe to its me topic to trigger `topicInit()`
- close connection for each user

I'm running the test locally so I modified `UserGet()`, `SubsForTopic()`, & `SubsForUser()` in rethinkdb adapter to have 100ms latency. You could see the changes here:
- https://github.com/riandyrn/chat/commit/6b7b0f771c326a6973a9089e895376dd60cb952e (same goroutine)
- https://github.com/riandyrn/chat/commit/744d3b713b0c8b4a3ab070e6b90d63e29bf9bc2d (different goroutine)

For testing engine I'm using [Tsung 1.7.0](http://tsung.erlang-projects.org/). Here is the repo for Tsung setup just in case you interested:
- https://github.com/riandyrn/tinode-user-generator/blob/master/tsung-topic-init-testing.xml

Here are the complete result of each test just in case you interested:
- http://riandyrn.github.io/benchmark-topic-init-same-goroutine/report.html (same goroutine)
- http://riandyrn.github.io/benchmark-topic-init-diff-goroutine/report.html (different goroutine)

So based on this result, I think it's better if we put `topicInit` on different goroutine.

What do you think, Gene?

or-else:
Thanks a lot, Riandy! Let's keep it on multiple go routines then. I'll discard this pull request.

PR #54, opened 2017-12-18
Fix Params Null in Response of {hi}

fatihkidstv:
Hello, Gene

I found glitch in `{hi}` response. The field `"params"` would be still printed even if it has zero value.

For details you could take a look at this snippet which demonstrate this issue: https://play.golang.org/p/lJiwbC8Ebg

Let me know your opinion.

Thanks


riandyrn:
Lol.., sorry it was my another account. I will create another PR.

PR #100, opened 2018-06-06
Add MYSQL DSN Environment Variable

TaylorChristie:
Currently, the MySQL database runs as the root user without a password, obviously that isn't the best for a live environment so I modified the docker files slightly to support a user defined DSN.


or-else:
Thanks!

TaylorChristie:
Could you push this change to the docker hub so I can start using it on the official image?

Thanks!

or-else:
I'm currently travelling thus delayed responses. Will try to do it over the weekend.

or-else:
done

PR #130, opened 2018-07-30
CMD instead of ENTRYPOINT

magicalbanana:
:man: :bear: :pig: 

Let's make it easier to run `docker run` without having to flag `--entrypoint` to override the entry point.


or-else:
Thanks!

PR #131, opened 2018-07-31
Do the same for tinode/chat

magicalbanana:
:man: :bear: :pig: 

Forgot to do this to the other Dockerfile.

### DIFF
--- a/docker/tinode/Dockerfile
+++ b/docker/tinode/Dockerfile
@@ -66,7 +66,7 @@ RUN chmod +x entrypoint.sh
 RUN chmod +x credentials.sh
 
 # Generate config from template and run the server.
-ENTRYPOINT ./entrypoint.sh
+CMD ./entrypoint.sh
 
 # HTTP and gRPC ports
 EXPOSE 18080 16061


or-else:
This is different from the chatbot's Dockerfile. Chatbot's ENTRYPOINT/CMD was a one-liner.
On the other hand  Tinode's `entrypoint.sh` contains a lot of logic: it generates a config file from the environment, optionally initializes the database, saves password for the chatbot. What's the value of not running the `entypoint.sh`? Tinode server won't fun without the config file. It's not possible to replace `entrypoit.sh` functionality with command line parameters.

magicalbanana:
@or-else I understand. The point of this is to be able to peer into the Docker image when it's running. Since the execution of the image is blocking we can't peer into the logs and what not. 

Basically, you can't do this with docker-compose:

```
docker-compose run --rm tinode /bin/bash
```

or-else:
But something like 
```
docker exec -it <container_name> /bin/bash
```
should work just fine.

The only difference I know between ENTRYPOINT and CMD is that CMD can be replaced with the command line parameters when starting a container while ENTRYPOINT cannot. Everything else is the same. It's just does not make sense not to run `entrypoint.sh` when starting a container. Maybe I'm missing something?

or-else:
I think this is no longer relevant. If not, please ping me.

PR #155, opened 2018-09-19
Add chatbot .net/.netcore implement 

BenDerPan:
Change Logs:
1. Add .net/.netcore (C#) version chatbot, I write it as the python chatbot. [more info](https://github.com/BenDerPan/chat/blob/master/chatbot/CSharp/readme.md)

2. folders change, I put the old python chatbot to "chatbot/Python",and the C# version to "chatbot/CSharp"

I tested it , it can work correctly, and when the server is down , it can be alright when the server is online again.

### DIFF
--- a/chatbot/CSharp/readme.md
+++ b/chatbot/CSharp/readme.md
@@ -1,0 +1,143 @@
+# Tinode Chatbot Example for .Net or .NetCore
+
+This is a rudimentary chatbot for Tinode using [gRPC API](../pbx/). It's written in C# as a demonstration
+that the API is language-independent.
+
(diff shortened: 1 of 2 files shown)


or-else:
Instead of moving python files to ./Python, please git-move them, i.e. 

`git mv * ./python/`

Git-moving preserves the commit history. Just moving without git erases the commit history.

or-else:
Please use all lowercase paths at the top level, i.e. `chatbot/python` and `chatbot/csharp`. It's an expectation and a common practice on Unix-like systems. It's fine to have upper case paths and names inside `chatbot/csharp` since it's Windows-specific.

Thanks!

BenDerPan:
@or-else git mv * ./python/ didn't work, it throw error: fatal: bad source, source=chatbot/*, destination=chatbot/python/*

PR #156, opened 2018-09-19
Add .net/.netcore implement chatbot

BenDerPan:
Change Logs:
1. Add .net/.netcore (C#) version chatbot, I write it as the python chatbot. [more info](https://github.com/BenDerPan/chat/blob/master/chatbot/csharp/readme.md)

2. folders change, I put the old python chatbot to "chatbot/python",and the C# version to "chatbot/csharp"

I tested it , it can work correctly, and when the server is down , it can be alright when the server is online again.

Fixed:
1. Git-moving preserves the commit history. Just moving without git erases the commit history.
2. Use all lowercase paths at the top level.


or-else:
Thanks!

I merged ur code to `devel`: https://github.com/tinode/chat/tree/devel/chatbot/csharp

I'm going to close this PR.

PR #160, opened 2018-10-02
Add new features and bug fixed, also published to nuget [Tinode.ChatBot]

BenDerPan:
ChangeLogs:
1. Add message  event notify, such as ServerPresEvent,ServerMetaEvent,CtrlMessageEvent,ServerDataEvent;
2. Add Subscriber for subscriber list;
3 Fixed no avatar meta data null reference exception;

I published the ChaBot .net library to [nuget.org](https://www.nuget.org/packages/Tinode.ChatBot/)

### DIFF
--- a/chatbot/csharp/Tinode.ChatBot.DemoNet46/Program.cs
+++ b/chatbot/csharp/Tinode.ChatBot.DemoNet46/Program.cs
@@ -1,5 +1,6 @@
 ﻿using CommandLine;
 using Google.Protobuf;
+using Newtonsoft.Json;
 using Pbx;
 using System;
 using System.Collections.Generic;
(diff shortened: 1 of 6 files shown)


or-else:
BTW, what do you think if I move your chatbot to a separate repository and give you commit access to it?

or-else:
Something like https://github.com/tinode/csharpbot. Or you come up with a better name for it.

BenDerPan:
That would be fine, your advised name is fine, but the release lib name will keep Tinode.ChatBot on nuget.

or-else:
Done. Once you accept the invite here https://github.com/tinode/csharpbot/invitations you should have commit access to the https://github.com/tinode/csharpbot

PR #168, opened 2018-10-09
Update data.json with anon access example

fstn:


fstn:
Add example that show how to give access to a topics for anon users

tinode-db/data.json
```diff
@@ -244,6 +245,14 @@
...
    "topic": "*BF",
    "user": "bob"
+  },
+  {
+   "createdAt": "-126h",
+   "name": "support",
+   "owner": "frank",
+   "tags": ["support","public"],
+   "public": {"fn": "Support", "photo": "support-64.jpg", "type": "jpg"},
+   "access": {"anon": 64, "auth": 64}
```
or-else: This is a good idea but tinode-db/makedb.go must be updated too. Otherwise it's not going to be used.

or-else:
What's the license of support.jpeg? Thanks!

fstn:
https://pixabay.com/fr/adulte-d-affaires-homme-d-affaires-1260380/

tinode-db/data.json
fstn: I commited changes to go file... but I don't know how to easily run it to test... what cmd line do you use to build and test makedb?

PR #192, opened 2018-11-29
Allow disabling Android notifications in FCM push messages

tiagoalves:
As discussed at: https://groups.google.com/forum/#!topic/tinode/KRQBB1r1OYk

This PR allows disabling the "Notification" payload for FCM push messages to Android. 

Motivation: Notification push messages are handled by the OS directly without any intervention by the mobile app. It's thus not possible to alter the notification such as using the sender's name or marking the message immediately as received.

Backwards compatibility was kept. The default continues to be that push messages are "data + notification".


or-else:
I'm at JFK waiting for my flight, will review after getting home.

tiagoalves:
Of course. No rush from my side.

PR #204, opened 2019-01-11
return authlvl when create user without login, which is used for root…

syklevin:
… user send msg  in pair with


or-else:
Thanks.

PR #260, opened 2019-08-16
Update API.md

Googlom:
seq -> seq_id


Googlom:
Hi. I was testing gRPC with BloomRPC UI client. In model.proto the field is `seq_id`. For example:
[https://github.com/tinode/chat/blob/2665ccb939e73f4191a0a7ac1c15f4b46c3751fa/pbx/model.proto#L299](url)
So, when I've imported .proto file to BloomRPC field was `seq` what gave me an error. I thought it is a typo in documentation.

or-else:
Yes, there are some differences in field names between protobuff and json APIs. The API.md is intended as a reference for json protocol. I don't see much value in a protobuff doc separate from comments in `model.proto`. If you wish, you may add more comments there. 

Thanks.

PR #261, opened 2019-08-20
fix for tinode#188

jimmykuu:
I reappear the problem, and I fixed it.


or-else:
Great catch! Thank you!

PR #274, opened 2019-10-01
Clustering support for user cache

or-else:
User cache is needed for iOS badge with the count of unread messages. It was implemented without regard for clustering. This PR adds clustering support to user cache.

### DIFF
--- a/server/cluster.go
+++ b/server/cluster.go
@@ -384,8 +384,9 @@ func (c *Cluster) Route(msg *ClusterReq, rejected *bool) error {
 // The Proxy expects no payload to be returned by the master.
 
 // UserCacheUpdate endpoint receives updates to user's cached values as well as sends push notifications.
-func (c *Cluster) UserCacheUpdate(req *UserCacheReq, rejected *bool) error {
-	usersRequestFromCluster(req)
+func (c *Cluster) UserCacheUpdate(msg *UserCacheReq, rejected *bool) error {
+	log.Printf("cluster: node '%s' received user cache update from node '%s'", c.thisNodeName, msg.Node)
+	usersRequestFromCluster(msg)
 	return nil
 }
 
@@ -404,10 +405,11 @@ func (c *Cluster) routeUserReq(req *UserCacheReq) error {
 			}
 			r := reqByNode[n.name]
 			if r == nil {
-				r = &UserCacheReq{PushRcpt: &push.Receipt{
-					Payload: req.PushRcpt.Payload,
-					To:      make(map[types.Uid]push.Recipient),
-				}}
+				r = &UserCacheReq{
+					PushRcpt: &push.Receipt{
+						Payload: req.PushRcpt.Payload,
+						To:      make(map[types.Uid]push.Recipient)},
+					Node: c.thisNodeName}
 			}
 			r.PushRcpt.To[uid] = recepient
 			reqByNode[n.name] = r
@@ -421,7 +423,7 @@ func (c *Cluster) routeUserReq(req *UserCacheReq) error {
 			}
 			r := reqByNode[n.name]
 			if r == nil {
-				r = &UserCacheReq{Inc: req.Inc}
+				r = &UserCacheReq{Node: c.thisNodeName, Inc: req.Inc}
 			}
 			r.UserIdList = append(r.UserIdList, uid)
 			reqByNode[n.name] = r
@@ -448,6 +450,7 @@ func (c *Cluster) routeUserReq(req *UserCacheReq) error {
 	if n == nil {
 		return errors.New("attempt to update user at a non-existent node (3)")
 	}
+	req.Node = c.thisNodeName
 	var rejected bool
 	err := n.call("Cluster.UserCacheUpdate", req, &rejected)
 	if rejected {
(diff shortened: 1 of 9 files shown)


server/user.go
```diff
@@ -566,19 +570,60 @@ func usersShutdown() {
...
 func usersUpdateUnread(uid types.Uid, val int, inc bool) {
-	if globals.usersUpdate != nil && (!inc || val != 0) {
+	if globals.usersUpdate == nil || (val == 0 && inc) {
+		return
+	}
+
+	upd := &UserCacheReq{UserId: uid, Unread: val, Inc: inc}
+	if globals.cluster.isRemoteTopic(uid.UserId()) {
+		// Send request to remote node which owns the user.
+		globals.cluster.routeUserReq(upd)
```
golangcibot: Error return value of `globals.cluster.routeUserReq` is not checked (from `errcheck`)

server/user.go
```diff
@@ -566,19 +570,60 @@ func usersShutdown() {
...
+		for uid, recepient := range rcpt.To {
+			if globals.cluster.isRemoteTopic(uid.UserId()) {
+				remote.PushRcpt.To[uid] = recepient
+			} else {
+				local.PushRcpt.To[uid] = recepient
+			}
+		}
+
+		if len(remote.PushRcpt.To) > 0 {
+			globals.cluster.routeUserReq(remote)
```
golangcibot: Error return value of `globals.cluster.routeUserReq` is not checked (from `errcheck`)

server/user.go
```diff
@@ -590,17 +635,23 @@ func usersRegisterUser(uid types.Uid, add bool) {
...
-	upd.uidList[0] = uid
+	upd := &UserCacheReq{UserIdList: make([]types.Uid, 1), Inc: add}
+	upd.UserIdList[0] = uid
 
-	select {
-	case globals.usersUpdate <- upd:
-	default:
+	if globals.cluster.isRemoteTopic(uid.UserId()) {
+		// Send request to remote node which owns the user.
+		globals.cluster.routeUserReq(upd)
```
golangcibot: Error return value of `globals.cluster.routeUserReq` is not checked (from `errcheck`)

server/run-cluster.sh
```diff
@@ -3,36 +3,48 @@
...
-      rm "/var/tmp/tinode-${node_name}.pid"
+      # Reda PIDs of running nodes from temp files and kill them.
+      kill `cat /var/tmp/tinode-${NODE_NAME}.pid`
+      # Clean up: delete temp files.
+      rm "/var/tmp/tinode-${NODE_NAME}.pid"
     done
     ;;
   *)
-    echo $"Usage: $0 {start|stop}"
+    echo $"Usage: $0 {start|stop} [config=./tinode.conf]"
```
aforge: Should it be
$ [CONFIG=<path_to_tinode.conf>] $0 {start|stop}
?

server/cluster.go
```diff
@@ -379,6 +380,85 @@ func (c *Cluster) Route(msg *ClusterReq, rejected *bool) error {
...
+}
+
+// Sends user cache update to user's Master node where the cache actually resides.
+// The request is extected to contain users who reside at remote nodes only.
+func (c *Cluster) routeUserReq(req *UserCacheReq) error {
+	// Index requests by cluster node.
+	reqByNode := make(map[string]*UserCacheReq)
+
+	if req.PushRcpt != nil {
+		// Request to send push notifications. Cretae separate packets for each affected cluster node.
```
aforge: s/Cretae/Create

server/user.go
```diff
@@ -566,19 +570,60 @@ func usersShutdown() {
...
 func usersPush(rcpt *push.Receipt) {
-	if globals.usersUpdate != nil {
+	if globals.usersUpdate == nil {
+		return
+	}
+
+	var local *UserCacheReq
+
+	// In case of a cluster pushes will be initiated at the nodes which own the users.
+	// Sort users into loacl and remote.
```
aforge: s/loacl/local

server/user.go
```diff
@@ -528,18 +528,22 @@ func replyDelUser(s *Session, msg *ClientComMessage) {
...
-	uid types.Uid
-	// User IDs being updated
-	uidList []types.Uid
+// UserCacheReq contains data which mutates one or more user cache entries.
+type UserCacheReq struct {
+	// Name of the node sending this request in case of cluster. Not set otherwise.
+	Node string
+
+	// UserId is set when a single user is updated.
+	UserId types.Uid
```
aforge: Why not use UserIdList[0] in such a case?

or-else: It's two different requests. `UserIdList` is used to increment or decrement count of topics a user is subscribed to. `UserId` is used to increment or decrement the count of unread messages for a give user. It's possible to use `UserIdList[0]` instead of `UserId` but then I would still have to use some heuristics to separate these use cases.

server/run-cluster.sh
or-else: Changed to `config` everywhere and updated usage.

server/cluster.go
or-else: done

server/user.go
or-else: done

server/user.go
aforge: Makes sense. Can you update the comments to make it explicit please?

server/run-cluster.sh
aforge: I meant "config" is an env var here. But the syntax is for a command line arg.
Or am I missing something here?

PR #277, opened 2019-10-04
Monitoring

or-else:
Quick and dirty exporter for Prometheus. It converts `expvar` to prometheus format.


monitoring/prometheus/main.go
```diff
@@ -0,0 +1,63 @@
...
+				promhttp.HandlerOpts{
+					ErrorLog: &promHTTPLogger{},
+				},
+			),
+		),
+	)
+
+	// Index page at web root.
+	http.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
+		w.Write([]byte(`<html><head><title>Tinode Exporter</title></head><body>
```
golangcibot: Error return value of `w.Write` is not checked (from `errcheck`)

monitoring/prometheus/exporter.go
```diff
@@ -0,0 +1,194 @@
...
+	}
+
+	ch <- prometheus.MustNewConstMetric(e.up, prometheus.GaugeValue, up)
+}
+
+func (e *Exporter) parseStats(ch chan<- prometheus.Metric, stats map[string]interface{}) error {
+
+	err := firstError(
+		e.parseAndUpdate(ch, e.version, prometheus.GaugeValue, stats, "Version"),
+		e.parseAndUpdate(ch, e.uptime, prometheus.CounterValue, stats, "uptime"),
```
aforge: I don't see these metrics in the expvar. Where are these coming from?

or-else: `Version` is added as a part of this PR: https://github.com/tinode/chat/pull/277/files#diff-d3d4b092ce8121dc129c8b29e3545206R549

`uptime` was left for reference. I've removed it now.

PR #284, opened 2019-10-13
{pres term} on rehashing to displaced topics (attempt 2).

aforge:
I've updated the comments. Let's agree on the data structure and I'll make appropriate changes in this PR.

### DIFF
--- a/server/cluster.go
+++ b/server/cluster.go
@@ -536,11 +536,10 @@ func (c *Cluster) routeToTopic(msg *ClientComMessage, topic string, sess *Sessio
 		return errors.New("attempt to route to non-existent node")
 	}
 
-	// Save node name: it's needed in order to inform relevant nodes when the session is disconnected.
-	if sess.nodes == nil {
-		sess.nodes = make(map[string]bool)
+	if sess.getRemoteSub(topic) == nil {
+		log.Println("Remote subscription missing for topic '%s', sid '%s'", topic, sess.sid)
+		sess.addRemoteSub(topic, &RemoteSubscription{node: n.name})
 	}
-	sess.nodes[n.name] = true
 
 	req := &ClusterReq{
 		Node:        c.thisNodeName,
@@ -595,11 +594,20 @@ func (c *Cluster) sessionGone(sess *Session) error {
 		return nil
 	}
 
-	// Save node name: it's need in order to inform relevant nodes when the session is disconnected
-	for name := range sess.nodes {
-		n := c.nodes[name]
+	notifiedNodes := make(map[string]bool)
+
+	sess.remoteSubsLock.Lock()
+	defer sess.remoteSubsLock.Unlock()
+
+	for _, remSub := range sess.remoteSubs {
+		nodeName := remSub.node
+		if notifiedNodes[nodeName] {
+			continue
+		}
+		notifiedNodes[nodeName] = true
+		n := c.nodes[nodeName]
 		if n != nil {
-			return n.forward(
+			if err := n.forward(
 				&ClusterReq{
 					Node:        c.thisNodeName,
 					Fingerprint: c.fingerprint,
@@ -609,7 +617,9 @@ func (c *Cluster) sessionGone(sess *Session) error {
 						RemoteAddr: sess.remoteAddr,
 						UserAgent:  sess.userAgent,
 						Ver:        sess.ver,
-						Sid:        sess.sid}})
+						Sid:        sess.sid}}); err != nil {
+				log.Printf("cluster: remote session shutdown failure: node '%s', error: '%s'", nodeName, err)
+			}
 		}
 	}
 	return nil
@@ -826,7 +836,9 @@ func (c *Cluster) invalidateRemoteSubs() {
 		var topicsToTerminate []string
 		for topic, remSub := range s.remoteSubs {
 			if remSub.node != c.ring.Get(topic) {
-				topicsToTerminate = append(topicsToTerminate, remSub.originalTopic)
+				if remSub.originalTopic != "" {
+					topicsToTerminate = append(topicsToTerminate, remSub.originalTopic)
+				}
 				delete(s.remoteSubs, topic)
 			}
 		}
--- a/server/cluster.go
+++ b/server/cluster.go
@@ -814,21 +814,19 @@ func (c *Cluster) rehash(nodes []string) []string {
 // Iterates over sessions hosted on this node and for each session
 // sends "{pres term}" to all displayed topics.
 // Called immediately after Cluster.rehash().
-func (c *Cluster) invalidateRemoteSubs(ss *SessionStore) {
-	ss.lock.Lock()
-	defer ss.lock.Unlock()
+func (c *Cluster) invalidateRemoteSubs() {
+	globals.sessionStore.lock.Lock()
+	defer globals.sessionStore.lock.Unlock()
 
-	for _, s := range ss.sessCache {
+	for _, s := range globals.sessionStore.sessCache {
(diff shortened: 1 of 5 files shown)


server/cluster.go
```diff
@@ -503,6 +503,16 @@ func (c *Cluster) isRemoteTopic(topic string) bool {
 	return c.ring.Get(topic) != c.thisNodeName
 }
 
+// Returns remote node name where the topic is hosted.
+// If the topic is hosted locally, returns an empty string.
+func (c *Cluster) nodeNameForTopicIfRemote(topic string) string {
+	key := c.ring.Get(topic)
```
or-else: I'm pretty sure it's going to crash when run standalone because `c` is `nil`.

server/cluster.go
```diff
@@ -796,3 +806,31 @@ func (c *Cluster) rehash(nodes []string) []string {
...
+		s.remoteSubsLock.Lock()
+		var topicsToTerminate []string
+		var keysToDelete []string
+		for topic, remSub := range s.remoteSubs {
+			if remSub.node != c.ring.Get(topic) {
+				topicsToTerminate = append(topicsToTerminate, remSub.originalTopic)
+				keysToDelete = append(keysToDelete, topic)
+			}
+		}
+		for _, topic := range keysToDelete {
```
or-else: I believe this is unnecessary. You can remove map entries inside the `range s.remoteSubs`. It's safe.

or-else: There is no value in moving it into a separate loop. It's still under the same loop.

server/cluster_leader.go
```diff
@@ -179,6 +179,7 @@ func (c *Cluster) sendPings() {
 
 		c.fo.activeNodes = activeNodes
 		c.rehash(activeNodes)
+		c.invalidateRemoteSubs(globals.sessionStore)
```
or-else: It does not make a ton of sense to pass global var `globals.sessionStore` as parameter. Unless you want to get rid of globals completely. which is not going to be trivial.

server/cluster.go
aforge: For sure. Fixed.

server/pres.go
```diff
@@ -295,6 +295,17 @@ func (t *Topic) presSubsOnlineDirect(what string) {
 	}
 }
 
+// Communicates "topic unaccessible (cluster rehashing or node connection lost)" event
+// to a list of topics promting the client to resubscribe to the topics.
+func (s *Session) presTermDirect(subs []string) {
+	log.Printf("sid '%s', uid '%s', terminating %s", s.sid, s.uid, subs)
+	msg := &ServerComMessage{Pres: &MsgServerPres{Topic: "me", What: "term"}}
+	for _, topic := range subs {
+		msg.Pres.Src = topic
```
or-else: This is likely to generate wrong topic names for p2p topics.

server/cluster.go

server/session.go
```diff
@@ -103,6 +111,11 @@ type Session struct {
 	// subs concurrently.
 	subsLock sync.RWMutex
 
+	// Map of remote topic subscriptions, indexed by topic name.
+	remoteSubs map[string]*RemoteSubscription
+	// Synchronizes access to remoteSubs.
+	remoteSubsLock sync.RWMutex
+
 	// Cluster nodes to inform when the session is disconnected
 	nodes map[string]bool
```
or-else: As I said in the previous review, this will likely contain stale entries. It's not horrible, just inefficient.

server/cluster.go
```diff
@@ -796,3 +810,29 @@ func (c *Cluster) rehash(nodes []string) []string {
...
+		if s.proto == CLUSTER || len(s.remoteSubs) == 0 {
+			continue
+		}
+		s.remoteSubsLock.Lock()
+		var topicsToTerminate []string
+		var keysToDelete []string
+		for topic, remSub := range s.remoteSubs {
+			if remSub.node != c.ring.Get(topic) {
+				topicsToTerminate = append(topicsToTerminate, remSub.originalTopic)
+				keysToDelete = append(keysToDelete, topic)
```
golangcibot: SA4010: this result of append is never used, except maybe in other appends (from `staticcheck`)

or-else: The bot is right. `keysToDelete` is not needed at all.

server/cluster.go
or-else: The bot complaining about `keysToDelete` is correct.

server/cluster_leader.go
aforge: Done. Using the global now.

server/pres.go
aforge: That's why I'm adding remSub.originalTopic to the array which gets passed to this function.
Let me double-check it.

server/cluster_leader.go
or-else: I don't see the change

server/cluster.go
aforge: Removed.

server/cluster.go
aforge: Removed it.

server/cluster_leader.go
aforge: Just did.

server/pres.go
aforge: Just re-verified: it does generate short names {"pres":{"topic":"me","src":"usrj_PEbl0quUg","what":"term"}}

or-else: Thanks!
What about new group topics, i.e. cases when the original topic name is like `new123`?

server/sessionstore.go
```diff
@@ -63,6 +63,7 @@ func (ss *SessionStore) NewSession(conn interface{}, sid string) (*Session, int)
 		s.send = make(chan interface{}, 256) // buffered
 		s.stop = make(chan interface{}, 1)   // Buffered by 1 just to make it non-blocking
 		s.detach = make(chan string, 64)     // buffered
+		s.remoteSubs = make(map[string]*RemoteSubscription)
```
or-else: It would be a good idea to initialize`s.remoteSubs` only when `globals.cluster` is not `nil`. Otherwise it's a waste of memory.

server/pres.go
aforge: Good catch! Changed the code to use expanded topic name.

server/sessionstore.go
aforge: Done.

server/session.go
or-else: Maybe get rid of it in favor of the `remoteSubs`

aforge: Just uploaded another commit. Please, take a look.

server/cluster.go
```diff
@@ -522,11 +536,10 @@ func (c *Cluster) routeToTopic(msg *ClientComMessage, topic string, sess *Sessio
 		return errors.New("attempt to route to non-existent node")
 	}
 
-	// Save node name: it's needed in order to inform relevant nodes when the session is disconnected.
-	if sess.nodes == nil {
-		sess.nodes = make(map[string]bool)
+	if sess.getRemoteSub(topic) == nil {
+		log.Println("Remote subscription missing for topic '%s', sid '%s'", topic, sess.sid)
```
golangcibot: printf: Println call has possible formatting directive %s (from `govet`)

server/session.go
or-else: LGTM

server/cluster.go
or-else: Yep, it's a legit one

golangcibot: printf: Println call has possible formatting directive %s (from `govet`)

PR #287, opened 2019-10-15
just a few typos

Googlom:
arg names according to implementation in **mysql/adapter.go** and **rethinkdb/adapter.go**


or-else:
Thnaks.

PR #289, opened 2019-10-15
duplicate of method CredGetActive in rethinkdb adapter

Googlom:
I don't know why but there is call to credGetActive from CredGetActive. I think its redundant.


or-else:
The reason why it `CredConfirm` calls `credGetActive` instead of `CredGetActive` is that I generally prefer exported adapter methods not to call other exported methods of the same adapter.
It ensures that there are no unintended consequences of changing an exported method. I would rather leave it as is. It's a small redundancy but it also serves a purpose of keeping `CredGetActive` and `CredConfirm` expicitly separate.

PR #306, opened 2019-10-29
Mongodb adapter branch (in development)

Googlom:
Almost 80% of the adapter has already been done, only some complex (for me) methods and creating indexes remain. I didn't use the context (context.Background() everywhere), I just didn't figure out how to use them here.


server/db/mongodb/adapter.go
```diff
@@ -0,0 +1,1521 @@
...
+
+	var err error
+	var config configType
+	if err = json.Unmarshal(jsonconfig, &config); err != nil {
+		return errors.New("adapter mongodb failed to parse config: " + err.Error())
+	}
+
+	var opts mdbopts.ClientOptions
+
+	if config.Addresses == nil {
```
or-else: You probably want to add login/password as configurable options.

or-else: Please add a .md file describing your storage schema, like https://github.com/tinode/chat/blob/master/server/db/rethinkdb/schema.md or https://github.com/tinode/chat/blob/master/server/db/mysql/schema.sql

or-else:
Please add a .md file describing your storage schema, like https://github.com/tinode/chat/blob/master/server/db/rethinkdb/schema.md or https://github.com/tinode/chat/blob/master/server/db/mysql/schema.sql

or-else:
> only some complex (for me) methods and creating indexes remain

If you need help please ask.

Googlom:
> Please add a .md file describing your storage schema, like https://github.com/tinode/chat/blob/master/server/db/rethinkdb/schema.md or https://github.com/tinode/chat/blob/master/server/db/mysql/schema.sql

server/db/mongodb/adapter.go
```diff
@@ -84,14 +85,14 @@ func (a *adapter) Open(jsonconfig json.RawMessage) error {
 		a.maxResults = defaultMaxResults
 	}
 
-	a.conn, err = mdb.Connect(ctx, &opts)
+	a.conn, err = mdb.Connect(a.ctx, &opts)
```
or-else: Context `a.ctx` is nil at this point. You initialize it below.

Googlom:
Do you have any scripts that somehow help with testing adapter methods? Unit tests or something like this?
For now I'm doing manual testing every time I finish a method and it's taking time...

or-else:
> Do you have any scripts that somehow help with testing adapter methods? Unit tests or something like this?
> For now I'm doing manual testing every time I finish a method and it's taking time...

No, unfortunately we do not. You are welcome to create them. 
You can run https://github.com/tinode/chat/blob/master/tn-cli/sample-script.txt as a kind of automated test for some of the methods.

Googlom:
I need a help with understanding compound indexes in RethinkDB. 
https://github.com/tinode/chat/blob/0428153dfd84323308a3de1552d922b0d6f14557/server/db/rethinkdb/adapter.go#L288-L294
Here the index contains references to documents sorted first by the values of the `Topic` field and, within each value of the `Topic` field, sorted by values of the `SeqId` field. Do I get it right?

But I didn't understand how that index works:
https://github.com/tinode/chat/blob/0428153dfd84323308a3de1552d922b0d6f14557/server/db/rethinkdb/adapter.go#L239-L248

or-else:
> Here the index contains references to documents sorted first by the values of the Topic field and, within each value of the Topic field, sorted by values of the SeqId field. Do I get it right?

Correct.

> But I don't understand how that index works:

The field `Devices` is an object with dynamically created fields like `51032a07a06ce13f`:

```
"Devices": {
  "51032a07a06ce13f": {
    "DeviceId": "dZ_erg687zY:APA91bHqjf6CUbbgZJVrdmO3sxcQ0Tx9xOmM2AwAo4gFthMPvDTghcxcFD8q6E74BiE89f3rnY12RTVTIj0w6ulRhc1ytH-NHzEJwCrduCGZsR7MaYMYD6RC_f8Q7zHFEb6w1nMYiJ2A" ,
   "Lang": "en_US" ,
    "LastSeen": Wed Oct 30 2019 06:00:02 GMT+00:00 ,
    "Platform": "android"
} ,
"5df48acc8383130a": {
  "DeviceId": "cLNdK_hI64M:APA91bFz1RQgTo0_PR59VKUHuwEvvqR6dKCW5sodkiankfNbRGi7fomhs3tPBEvKRGbMmvsxoH2gmmDeRSRz7Ma2CJfeSAiYM3Uu2ePCrRxfQW1tB8W1GH2kzpjSmoMyMXmJ8AKbUWVI" ,
    "Lang": "" ,
    "LastSeen": Thu Oct 24 2019 06:35:30 GMT+00:00 ,
    "Platform": "android"
  } 
}
```

So this expression indexes `DeviceId` fields within these objects.

Googlom:
Seems mongo do not support this kind of indexing: [https://stackoverflow.com/questions/17134175/does-indexes-on-dynamic-nested-keys-works-in-mongodb](url)
So I can change scheme to something like:
```
  "devices": [
    {
      "_id": "5df48acc8383130a",
      "deviceid": "dZ_erg687zY:APA91bHqjf6CUbbgZJVrdmO3sxcQ0Tx9xOmM2AwAo4gFthMPvDTghcxcFD8q6E74BiE89f3rnY12RTVTIj0w6ulRhc1ytH-NHzEJwCrduCGZsR7MaYMYD6RC_f8Q7zHFEb6w1nMYiJ2A",
      "lang": "en_US",
      "lastseen": "2019-10-11T12:13:14.522Z",
      "platform": "android"
    },
     {
      "_id": "51032a07a06ce13f",
      "deviceid": "cLNdK_hI64M:APA91bFz1RQgTo0_PR59VKUHuwEvvqR6dKCW5sodkiankfNbRGi7fomhs3tPBEvKRGbMmvsxoH2gmmDeRSRz7Ma2CJfeSAiYM3Uu2ePCrRxfQW1tB8W1GH2kzpjSmoMyMXmJ8AKbUWVI",
      "lang": "",
      "lastseen":"2019-10-11T12:13:14.522Z",
      "platform": "android"
    }
  ]
```
and then create unique index on `devices.deviceid` or replace `_id` field with `deviceid` and then create index on `devices._id`. I think it won't affect the rest of code. (or will?)

or-else:
> and then create unique index on `devices.deviceid` or replace `_id` field with `deviceid` and then create index on `devices._id`. I think it won't affect the rest of code. (or will?)

Yes, it should be fine. Just make sure it's also easy for you to add and remove devices. I.e. add or remove elements from `devices` array.

Googlom:
All methods and indexes are done now. Adapter almost ready - should be heavily tested before it can be recommended for production use. All tests that I've done are manual, so there is no guarantees...

or-else:
One way of doing it is to make it clear that the adapter is experimental/not production tested and let users take the risks. 

Please also add installation instructions to https://github.com/tinode/chat/blob/master/INSTALL.md.

Then some minor edits in these files:
https://github.com/tinode/chat/blob/master/README.md
https://github.com/tinode/chat/blob/master/docker/tinode/Dockerfile
https://github.com/tinode/chat/blob/master/docker/README.md
https://github.com/tinode/chat/blob/master/tinode-db/README.md
https://github.com/tinode/chat/blob/master/chatbot/python/README.md

I'm going to update https://github.com/tinode/chat/blob/master/server/.golangci.yml in `devel` with `mongodb` buildtag. Once merged, it will catch some some typos in the comments in your PR. Please fix them.

Looks really nice. Thank you!

Googlom:
> Please also add installation instructions to https://github.com/tinode/chat/blob/master/INSTALL.md.
> 
> Then some minor edits in these files:
> https://github.com/tinode/chat/blob/master/README.md
> https://github.com/tinode/chat/blob/master/docker/tinode/Dockerfile
> https://github.com/tinode/chat/blob/master/docker/README.md
> https://github.com/tinode/chat/blob/master/tinode-db/README.md
> https://github.com/tinode/chat/blob/master/chatbot/python/README.md
> 

done 

> I'm going to update https://github.com/tinode/chat/blob/master/server/.golangci.yml in `devel` with `mongodb` buildtag. Once merged, it will catch some typos in the comments in your PR. Please fix them.

I don't quite understand what it's about.. sorry, new to Go :)

Googlom:
I want a suggestion on how to correctly write tests on adapter. Test `store` methods or directly `adapter` methods?
And there may be cases when there are specific requests to the database. For example, after running adapter.TopicDelete method I want to query DB to check if its really hard/soft deleted. Or after some updating method, check if fields are correctly updated.

or-else:
> I want a suggestion on how to correctly write tests on adapter. Test `store` methods or directly `adapter` methods?

Certainly adapter methods. Otherwise you would be testing two things at once and it's not the best practice.

> And there may be cases when there are specific requests to the database. For example, after running adapter.TopicDelete method I want to query DB to check if its really hard/soft deleted. Or after some updating method, check if fields are correctly updated.

I would first test creation of database, then fill it with dummy data, testing that insertions work, then execute tests of mutations.

or-else:
GolangCLI is a linter. It finds mistakes which are not quite errors. Like typos in comments, inefficient copy or assignments, etc. It runs automatically on every commit but in case of this PR it does not run because it does not know about `mongodb` build tag.

Googlom:
I'm having some problems while testing `adapter.MessageSave()` method. Because of `store.GetUid()`
https://github.com/tinode/chat/blob/f7816b8d69a823ecb556f74040f95b52339b3d95/server/db/rethinkdb/adapter.go#L1485-L1489
I don't initialize store, but adapter itself. So, `uGen` not initialized. Maybe `store.GetUid()` should be moved to store package? `store.GetUid()` used only twise inside **adapter** and once inside **store**. I'll create PR, so you can see changes

or-else:
Your proposal is reasonable in the context of testing mongo or rethink adapters but it would create problems with mysql adapter. MySQL adapter relies on `DecodeUid` and `EncodeUid`. They are defined right after the `GetUid`. Consequently I don't think it's a good idea to remove `GetUid` because it implies that `DecodeUid` and `EncodeUid` are independent. 

I see what you need and it's clear that it legit. Maybe UID manipulation should be moved to a separate package. 

I think it's reasonable to move `msg.SetUid()` from adapter to store. It should also solve your immediate problem. But please do not remove `GetUid()`.

or-else:
Is it ready to be merged into devel? If so, I can merge it and add mondo db to our release process.

Googlom:
Many bugs came out during unit tests. They were fixed but I think there is more. Let me test some more. I'll tell when it's ready to merge

Googlom:
Seems like everything works as expected. I think it's ready.
I'll fix bugs as they appear.

or-else:
Ask asked my colleague to join for the review. It might take a day or two. Thanks!

server/db/mongodb/adapter.go
```diff
@@ -0,0 +1,2159 @@
...
+	maxResults int
+	version    int
+	ctx        context.Context
+}
+
+const (
+	defaultHost     = "localhost:27017"
+	defaultDatabase = "tinode"
+
+	adpVersion  = 108
```
or-else: Please sync the version with the other databases (110).

server/db/mongodb/adapter.go
```diff
@@ -0,0 +1,2159 @@
...
+// TODO: UpgradeDb upgrades database to the current adapter version.
+func (a *adapter) UpgradeDb() error {
+	return nil
+}
+
+// Create system topic 'sys'.
+func createSystemTopic(a *adapter) error {
+	now := t.TimeNow()
+	_, err := a.db.Collection("topics").InsertOne(a.ctx, &t.Topic{
+		ObjHeader: t.ObjHeader{Id: "sys", CreatedAt: now, UpdatedAt: now},
```
or-else: A change was made in other adapters: `TouchedAt` is now required too.

server/db/mongodb/adapter.go
```diff
@@ -0,0 +1,2163 @@
...
+	if len(remove) > 0 {
+		newTags = diff(newTags, remove)
+	}
+
+	update := map[string]interface{}{"tags": newTags}
+	if err := a.UserUpdate(uid, update); err != nil {
+		return nil, err
+	}
+
+	// Get the new tags
```
aforge: Why do you need to fetch tags here?
Can you just return newTags?

aforge: Why do you need to fetch tags here?
Can you just return newTags?

server/db/mongodb/adapter.go
```diff
@@ -0,0 +1,2163 @@
...
+			"foreignField": "_id",
+			"as":           "fromTopics"},
+		},
+		// Merge two documents into one
+		b.M{"$replaceRoot": b.M{"newRoot": b.M{"$mergeObjects": b.A{b.M{"$arrayElemAt": b.A{"$fromTopics", 0}}, "$$ROOT"}}}},
+
+		b.M{"$match": b.M{
+			"deletedat": b.M{"$exists": false},
+			// Filter by access mode
+			"modewant":  b.M{"$bitsAllSet": b.A{1}},
```
aforge: b.A{1} - it reads you want bit 1 to be set.
Is the read permission not bit 0? Or am I missing something here?

server/db/mongodb/adapter.go
```diff
@@ -0,0 +1,2163 @@
...
+				return err
+			}
+			if err = a.AuthDelScheme(uid, scheme); err != nil {
+				return err
+			}
+			return sess.CommitTransaction(a.ctx)
+		}); err != nil {
+			return err
+		}
+		sess.EndSession(a.ctx)
```
aforge: defer?

server/db/mongodb/adapter.go
```diff
@@ -0,0 +1,2163 @@
...
+	if err := createSystemTopic(a); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// TODO: UpgradeDb upgrades database to the current adapter version.
+func (a *adapter) UpgradeDb() error {
+	return nil
```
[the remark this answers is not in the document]
or-else: Actually, I disagree. No error is needed here. The `nil` error is a valid response to NOOP `UpgradeDb`.

server/db/mongodb/adapter.go
```diff
@@ -0,0 +1,2163 @@
...
+		TouchedAt: now,
+		Access:    t.DefaultAccess{Auth: t.ModeNone, Anon: t.ModeNone},
+		Public:    map[string]interface{}{"fn": "System"},
+	})
+	return err
+}
+
+// User management
+
+// UserCreate creates user record
```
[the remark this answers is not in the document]
or-else: I also disagree. Go's official code style requires function name to be the first word in a comment.

server/db/mongodb/adapter.go
```diff
@@ -0,0 +1,2163 @@
...
+				return err
+			}
+		}
+
+		// Finally commit all changes
+		return sess.CommitTransaction(a.ctx)
+	}); err != nil {
+		return err
+	}
+	sess.EndSession(a.ctx)
```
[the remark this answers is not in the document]
or-else: This is the last statement in a function call. `Defer` would add no value.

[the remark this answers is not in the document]
or-else: Actually, scratch that.  `Defer` right after opening and checking for non-nil would make sense.

server/db/mongodb/adapter.go
or-else: `newTags` is really not new tags, but rather a mutation operation of old tags into new tags. So, the tags must be fetched either before or after the mutation. Fetching them after the mutation seems to be safer.

aforge: Ah, sure. Makes sense.
Can you add a comment explaining this please?

server/db/mongodb/adapter.go
```diff
@@ -0,0 +1,2163 @@
...
+	"github.com/tinode/chat/server/store"
+	t "github.com/tinode/chat/server/store/types"
+	b "go.mongodb.org/mongo-driver/bson"
+	"go.mongodb.org/mongo-driver/bson/primitive"
+	mdb "go.mongodb.org/mongo-driver/mongo"
+	mdbopts "go.mongodb.org/mongo-driver/mongo/options"
+)
+
+// adapter holds MongoDB connection data.
+type adapter struct {
```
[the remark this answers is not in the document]
Googlom: I don't understand what you're about.. I've tried to follow style and naming like in RethinkDB adapter. I don't see difference here...

server/db/mongodb/adapter.go
```diff
@@ -0,0 +1,2163 @@
...
+			if _, err = a.db.Collection("topics").UpdateMany(a.ctx, b.M{"_id": b.M{"$in": topicIds}}, disable); err != nil {
+				return err
+			}
+			if _, err = a.db.Collection("users").UpdateMany(a.ctx, b.M{"_id": uid.String()}, disable); err != nil {
+				return err
+			}
+		}
+
+		// Finally commit all changes
+		return sess.CommitTransaction(a.ctx)
```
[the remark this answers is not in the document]
Googlom: My fault. I've forgot about aborting transaction. Will fix it soon.

server/db/mongodb/adapter.go
Googlom: This `match` returns all documents where `modewant`s **bit-1** set to 1 (have read permission)

aforge: I see. Thanks for explaining.

server/db/mongodb/adapter.go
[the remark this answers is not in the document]
aforge: Sounds good. Please, check other places where you work with transactions for similar logic.

or-else:
Thanks!

Googlom:
👍

PR #311, opened 2019-11-14
Try to catch cases when remote subscription is missing

or-else:
This is a bit extreme but missing remote subscription is a bug. Crash should allow us to catch it if it ever happens.

### DIFF
--- a/server/cluster.go
+++ b/server/cluster.go
@@ -537,7 +537,7 @@ func (c *Cluster) routeToTopic(msg *ClientComMessage, topic string, sess *Sessio
 	}
 
 	if sess.getRemoteSub(topic) == nil {
-		log.Println("No remote subscription (yet) for topic '%s', sid '%s'", topic, sess.sid)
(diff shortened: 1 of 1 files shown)


or-else:
I first added panic then removed it: it's possible and ordinary to receive a message without a subscription. Adding a subscription for every message is likely to cause a memory leak. But that's probably not the cause of the problem.

server/cluster.go
```diff
@@ -537,8 +537,7 @@ func (c *Cluster) routeToTopic(msg *ClientComMessage, topic string, sess *Sessio
 	}
 
 	if sess.getRemoteSub(topic) == nil {
-		log.Printf("Remote subscription missing for topic '%s', sid '%s'", topic, sess.sid)
-		sess.addRemoteSub(topic, &RemoteSubscription{node: n.name})
+		log.Println("No remote subscription (yet) for topic '%s', sid '%s'", topic, sess.sid)
```
golangcibot: printf: Println call has possible formatting directive %s (from `govet`)

PR #313, opened 2019-11-15
move store.GetUid() from adapter to store

Googlom:
Issue: https://github.com/tinode/chat/pull/306#issuecomment-554343709


or-else:
As I said in the issue, please just move `msg.SetUid(store.GetUid())` from adapter to store, but do not remove `store.GetUid()`. Thanks!

Googlom:
Done

PR #323, opened 2019-11-27
Support for delayed presence notifications

or-else:
This is a rough draft, not thoroughly tested.

If client requests for notifications to be delayed they are placed into a queue. A ticker pulls notifications from the queue if they are old enough and executes them. If session disconnects from the topic before the notifications are sent, the request is removed from the queue.


server/topic.go
```diff
@@ -577,10 +591,33 @@ func (t *Topic) run(hub *Hub) {
...
+				if expiration.Before(sreg.pkt.timestamp) {
+					// All done. Remaining events are newer.
+					break
+				}
+				t.defrNotif.Remove(elem)
+				if pssd, ok := t.sessions[sreg.sess]; ok {
+					pssd.ref = nil
+					t.sessions[sreg.sess] = pssd
+				}
+				t.sendSubNotifications(types.ParseUserId(sreg.pkt.from), sreg)
```
aforge: Should we also update online counters?

server/topic.go
```diff
@@ -2645,41 +2704,27 @@ func (t *Topic) subsCount() int {
...
-		// All uids are removed.
-		delete(t.sessions, s)
-	} else if uids.Rem(user) {
-		// One uid is removed
-		if len(uids) > 0 {
-			t.sessions[s] = uids
-		} else {
-			delete(t.sessions, s)
+// Removes session record if 'asUid' matches subscribed user.
+func (t *Topic) remSession(s *Session, asUid types.Uid) *perSessionData {
```
aforge: Old implementation assumed multiple uids per session.
Was it wrong (i.e. this PR fixes it up)? If not, in which cases was that possible?

server/topic.go
```diff
@@ -875,22 +928,25 @@ func (t *Topic) subCommonReply(h *Hub, sreg *sessionJoin) error {
 
 // User requests or updates a self-subscription to a topic. Called as a
 // result of {sub} or {meta set=sub}.
+// Returns changed == true if user's accessmode has changed.
+//
+//	h			- hub
```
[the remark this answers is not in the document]
or-else: Actually it is consistent according to Golang specs: 4 chars per tab.  The issue is with Github rendering.
Screenshot 2019-11-28 at 10 35 24

server/topic.go
or-else: It never worked. Session always refused the second subscription to the same topic even if it was on behalf of a different user:
https://github.com/tinode/chat/blob/master/server/session.go#L410

I removed it in https://github.com/tinode/chat/pull/323/commits/e8839894b118ef72e214dadfee320532690320b2

This seems to be a very niche corner case (root user subscribes to the same topic from the same session on behalf of two different users). Supporting it would add too much complexity for such an unimportant feature.

server/push/fcm/push_fcm.go
```diff
@@ -182,6 +182,7 @@ func sendNotifications(rcpt *push.Receipt, config *configType) {
 					}
 					if config.IncludeAndroidNotification {
 						msg.Android.Notification = &fcm.AndroidNotification{
+							Tag:   rcpt.Payload.Topic,
```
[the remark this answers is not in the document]
or-else: `msg.Android.Notification` is the equivalent of iOS's notification with `ContentAvailable: false`. I.e. if the app is in the background and `msg.Android.Notification` is received the app will not be woken up. Tag allows grouping together notifications on the same topic.

server/topic.go
or-else: Online counters are updated here:
https://github.com/tinode/chat/blob/sync/server/topic.go#L1198
https://github.com/tinode/chat/blob/sync/server/topic.go#L239

And I added one more case where it needs to be updated.

server/topic.go
```diff
@@ -136,6 +140,14 @@ type perSubsData struct {
 	enabled bool
 }
 
+// Data related to a subscription of a sesstion to a topic.
```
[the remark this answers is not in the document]
or-else: done

server/topic.go
aforge: https://github.com/tinode/chat/blob/sync/server/topic.go#L1201 increments the counter for foreground sessions only when a session subscribes to the topic.

This piece of code sends presences notifications if the session was subscribed in the background but didn't go away within `deferredNotificationsTimeout` seconds.
So sending `{pres}` means we are promoting the session to foreground but the counter won't be incremented. Am I missing something?

server/topic.go
aforge: Makes sense. Thanks for explaining.

server/push/fcm/push_fcm.go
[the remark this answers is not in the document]
aforge: Can you add a short comment please?

server/topic.go
or-else: You are correct and I fixed it here: https://github.com/tinode/chat/pull/323/commits/4eaaa40c639471a8d937e0d0a8538bb0f19fef33

server/push/fcm/push_fcm.go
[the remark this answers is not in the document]
or-else: done

PR #327, opened 2019-12-01
Proposal for reporting deleted messages

or-else:
This is the proposal for dealing with

> Topic.subscribe fetches only the latest |limit| messages posted after latest cached message.
> Clients should be able to figure out legit gaps in sequence numbers and fetch the missing messages from the server when necessary.


docs/API.md
```diff
@@ -1038,13 +1038,18 @@ data: {
...
+              // of deleted messages; inclusive-exclusive, i.e. [low, hi), optional
   content: { ... } // object, application-defined content exactly as published
               // by the user in the {pub} message
 }
 ```
 
-Data messages have a `seq` field which holds a sequential numeric ID generated by the server. The IDs are guaranteed to be unique within a topic. IDs start from 1 and sequentially increment with every successful `{pub}`](#pub) message received by the topic.
+Data messages generally have a `seq` field which holds a sequential numeric ID generated by the server. The IDs are guaranteed to be unique within a topic. IDs start from 1 and sequentially increment with every successful [`{pub}`](#pub) message received by the topic.
+
+When user requests a range of messages by issuing a [`{get what=data}`](#get) request, the server responds by sending requested `{data}` messages. If any of the messages in range are deleted, the server also sends one or more `{data}` messages without the `seq` and `content` fields but with `delseq` field. The field contains an array of IDs of deleted message ranges.
```
aforge: Should we also honor `since`, `before` and `limit` params in `{get what=data}` messages?
It seems logical but OTOH since the purpose of this effort is letting the client figure out the gaps it may be more reasonable to send all the data all the time. I'm personally more inclined to the former. What do you think?

or-else: > Should we also honor since, before and limit params in {get what=data} messages?

They are honored:
https://github.com/tinode/chat/blob/master/server/db/mysql/adapter.go#L1886

or-else:
@aforge I've mostly added the changes to webapp and tindroid, but it bothers me a bit. 

This change creates redundancy. We already have `{del}` packet which tells the client which message ranges have been deleted. We partially replicate the same functionality using `{data}` packet. This approach simplifies client a bit at the expense of making the API less clear.

What do you think?

server/store/store.go
```diff
@@ -605,7 +605,63 @@ func (MessagesObjMapper) DeleteList(topic string, delID int, forUser types.Uid,
...
+	if since < 1 {
+		since = 1
+	}
+
+	minId := opt.Before
+	if count > 0 {
+		minId = messages[count-1].SeqId
+	}
+
+	// Fewer messages returned than requested: all messages are deleted between 'since' and the lowes seqId in 'messages'.
```
aforge: s/lowes/lowest

or-else:
I'm closing this per our conversation.

Issue #330, opened 2019-12-13
Is it possible to make mongodb replica_set configurable?


beanhacker:
Hello author,
    the MONOGODB I use here does not do a replication set. If I do a replication set at this time, many other services will also be forced to add the replication set option when connecting. Need to modify the system a lot. replica_set made configurable? thank!

aforge:
You mean you are running a standalone mongodb deployment?

or-else:
Hey @Googlom,

Please take a look at this request. Making replica set optional looks like a reasonable idea to me.

Googlom:
Transactions don't work without replicaset configured on MongoDB. I'll have to remove transaction support to be able to run on standalone MongoDB instance. Or duplicate query logic to support both transaction and non-transaction queries - then it'll be possible to do replica set optional.

or-else:
I think it's a tradeoff which the user should be able to make. We just need to make it clear to the user. Maybe even write it to the log when initalizing the adapter.

It's a actually quite easy to make transactions optional. Define something like 
```
func (a *adapter) maybeStartTransaction(sess mdb.Session) error {
  if a.useTransactions {
    return sess.StartTransaction()
  }
  return nil
}

func (a *adapter) maybeCommitTransaction(sess mdb.Session) error {
  ... same as above but with sess.CommitTransaction ...
}
```
And then use these calls everywhere instead of `sess.StartTransaction`/`sess.CommitTransaction`.

or-else:
@Googlom would you make the changes or should I?

Googlom:
I will make the changes tomorrow

or-else:
Thanks!

or-else:
Done in #333

PR #332, opened 2019-12-16
Make replicaset optional

Googlom:
fix to #330


Googlom:
missed with branch :D

PR #335, opened 2019-12-31
Fix for missing presence notifications

or-else:
Here is the fix.

### DIFF
--- a/server/topic.go
+++ b/server/topic.go
@@ -99,11 +99,9 @@ type Topic struct {
 	// Channel to terminate topic  -- either the topic is deleted or system is being shut down. Buffered = 1.
 	exit chan *shutDown
 
-	// Flag which tells topic to stop acception requests: hub is in the process of shutting it down
-	suspended atomicBool
+	// Flag which tells topic lifecycle status: ready, suspended, marked for deletion.
+	status int32
 }
-
-type atomicBool int32
 
 // perUserData holds topic's cache of per-subscriber data
 type perUserData struct {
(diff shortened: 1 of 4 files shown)


server/topic.go
```diff
@@ -2580,24 +2586,47 @@ func (t *Topic) mostRecentSession() *Session {
 	return sess
 }
 
-func (t *Topic) suspend() {
-	atomic.StoreInt32((*int32)(&t.suspended), 1)
+const (
+	// Topic is just created.
+	topicStatusNew = 0
+	// Topic is fully initialized.
+	topicStatusLoaded = 1
```
aforge: I understand `topicStatusLoaded` is only meaningful in ME, P2P and GRP contexts.
FND and SYS topics will forever remain in `topicStatusNew` state.
Is that the intent?
If so, perhaps, SYS and FND should go to `topicStatusLoaded` right away?

server/topic.go
```diff
@@ -782,39 +784,12 @@ func (t *Topic) handleSubscription(h *Hub, sreg *sessionJoin) error {
 	return nil
 }
 
-// Send presence notification in response to a subscription
-func (t *Topic) sendSubNotifications(asUid types.Uid, sreg *sessionJoin) {
+// Send immediate presence notification in response to a subscription.
+// These notifications are always sent immediately even if background is requested.
```
aforge: This defeats the purpose of introducing background requests.
We don't care about presence notifications when background a request is sent. Should we discuss this further?

or-else: I split the notifications into two sets: sent immediately regardless of the background status and those optionally delayed. The always immediate notifications are p2p & group notifications on a new subscription. I think it's a weird case to join a new chat in background.

server/topic.go
or-else: done

server/topic.go
aforge: Okay, thanks for explaining.
This is good.

PR #348, opened 2020-02-01
Ability to send different types of push notifications.

or-else:
This is needed in order to send pushes for new subscriptions as well as silencing pushes when all messages are read from another device.


server/push/fcm/push_fcm.go
```diff
@@ -111,27 +111,32 @@ func payloadToData(pl *push.Payload) (map[string]string, error) {
...
 
-	// Trim long strings to 80 runes.
-	// Check byte length first and don't waste time converting short strings.
-	if len(data["content"]) > 80 {
-		runes := []rune(data["content"])
-		if len(runes) > 80 {
-			data["content"] = string(runes[:80]) + "…"
+		// Trim long strings to 80 runes.
+		// Check byte length first and don't waste time converting short strings.
+		if len(data["content"]) > 80 {
```
aforge: Make 80 a constant?

or-else: done

PR #350, opened 2020-02-02
Enable history and command editing in interactive mode in tn-cli.

aforge:


tn-cli/tn-cli.py
```diff
@@ -24,6 +24,7 @@
     import queue
 import random
 import re
+import readline
```
or-else: Please add readline to requirements.txt.

or-else:
It looks like readline is not available on Windows: https://bugs.python.org/issue26258

tn-cli/tn-cli.py
or-else: There is something called pyreadline which exists on windows but seems to be abandoned:
https://pypi.org/project/pyreadline/

aforge: Good catch! Yes, I'll check pyreadline - we need to see if it's compatible with raw_input()/input().

or-else: Please also look at https://github.com/prompt-toolkit/python-prompt-toolkit
It seems to be well-supported albeit huge.

aforge:
Incorporated prompt toolkit into this PR.

or-else:
It it ready for review?

aforge:
Correct. Please, take a look.

or-else:
`python-prompt-toolkit` works fine but `tn-cli` in general is broken. Not because of this change, it's broken in devel as well.

PR #347, opened 2020-02-01
Ability to suspend users

or-else:
Only mysql and rethinkdb adapters are updated for new functionality.

PR #351, opened 2020-02-02
Ability to suspend users (MongoDB)

Googlom:
#347

### DIFF
--- a/server/db/mongodb/adapter.go
+++ b/server/db/mongodb/adapter.go
@@ -35,7 +35,7 @@ const (
 	defaultHost     = "localhost:27017"
 	defaultDatabase = "tinode"
 
-	adpVersion  = 110
+	adpVersion  = 111
 	adapterName = "mongodb"
 
 	defaultMaxResults = 1024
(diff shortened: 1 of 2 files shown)


or-else:
Please also update https://github.com/Googlom/chat/blob/suspend/server/db/mongodb/schema.md

or-else:
One more thing: https://github.com/tinode/chat/commit/979eb31ebc18a39857ed8dbce2675f6d5186f288
I missed this rethinkdb change in my earlier commits.

server/db/mongodb/adapter.go
```diff
@@ -1330,8 +1427,8 @@ func (a *adapter) TopicDelete(topic string, hard bool) error {
 	} else {
 		now := t.TimeNow()
 		_, err = a.db.Collection("topics").UpdateOne(a.ctx, filter, b.M{"$set": b.M{
-			"updatedat": now,
-			"deletedat": now,
+			"state":   t.StateDeleted,
```
aforge: Do you want to update `updatedat` here as well?

Googlom: https://github.com/tinode/chat/blob/1c74854863439fe5defa6d0c099be0d29cd4bf11/server/db/rethinkdb/adapter.go#L1244-L1249

`updatedat` is not updating in rethinkdb

or-else: Let me check if it needs to be changed too. Maybe I made a mistake. I'll get back soon.

or-else: Current code is correct. `updatedat` should NOT be updated on state change.

PR #353, opened 2020-02-05
Pushes for new subscriptions

or-else:
Also AndroidNotification refactored. This refactoring is not directly related to the new push type.

### DIFF
--- a/server/push/fcm/push_fcm.go
+++ b/server/push/fcm/push_fcm.go
@@ -120,10 +120,9 @@ func payloadToData(pl *push.Payload) (map[string]string, error) {
 	}
 	data["topic"] = pl.Topic
 	data["ts"] = pl.Timestamp.Format(time.RFC3339Nano)
-	// Must use "xfrom" because "from" is a reserved word.
-	// Google did not bother to document it anywhere.
+	// Must use "xfrom" because "from" is a reserved word. Google did not bother to document it anywhere.
 	data["xfrom"] = pl.From
-	if pl.SeqId > 0 {
+	if pl.What == push.ActMsg {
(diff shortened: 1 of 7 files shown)


server/push/fcm/push_fcm.go
```diff
@@ -37,15 +37,112 @@ type Handler struct {
 	client *fcm.Client
 }
 
+// Configuration of AndroidNotification payload.
+type androidConfig struct {
```
aforge: Can you move it closer to other struct definitions for better readability please?

server/push/fcm/push_fcm.go
```diff
@@ -175,6 +236,19 @@ func sendNotifications(rcpt *push.Receipt, config *configType) {
...
 	}
 
+	var titlelc, title, bodylc, body, icon, color string
+	if config.Android.Enabled {
+		titlelc = config.Android.getTitleLocKey(rcpt.Payload.What)
+		title = config.Android.getTitle(rcpt.Payload.What)
+		bodylc = config.Android.getBodyLocKey(rcpt.Payload.What)
+		body = config.Android.getBody(rcpt.Payload.What)
+		if body == "$content" {
+			body = data["content"]
```
aforge: Should we also check if `body` is empty here?

or-else: Maybe it's intentionally blank.

server/push/fcm/push_fcm.go
or-else: The logic I followed was type definition immediately followed by all the methods of the type, then the next type definition. I can do [type def 1], [type def 2]..., [methods for type 1], etc but I don't think it's objectively better, just a matter of preference.

PR #359, opened 2020-02-13
Add macro functionality to tn-cli.

aforge:
Macros are high-level commands that expand to a sequence of
basic Tinode CLI commands.

### DIFF
--- a/tn-cli/sample-macro-script.txt
+++ b/tn-cli/sample-macro-script.txt
@@ -1,0 +1,46 @@
+# This script shows what can be done with tn-cli macros. Run it as
+#
+#   python tn-cli.py --no-login < sample-macro-script.txt
+#
+# The script performs the following:
+#
+#  - Logs in as Xena (user has root privileges)
+#  - Creates a new user Test User with useradd
+#  - Changes Test User's name to 'Test User 1'
+#  - Modifies Test User's default anon acs to JR with chacs
+#  - Sets Test User's password to test456
+#  - Deletes Test User with userdel
+#  - Uses resolve to locate user Alice
+#  - Subscribes to Alice
+#  - Sends a message to Alice
+
+# Create user 'Test User'
+.must $user useradd --name 'Test User' --password test123 --comment test0 \
+  --cred email:test@example.com --avatar ./test-128.jpg --tags test,test-user \
+  --auth=JRWPA --anon=JW test
+
+# Login as xena.
+.must $xena login --scheme=basic --secret=xena:xena123
+
+# Change test's publis name.
+.must usermod $user.params[user] --name 'Test User 1'
+
+# Change test's anon acs.
+must chacs $user.params[user] --anon=JR
+
+# Set test's password to test456.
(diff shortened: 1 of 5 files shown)


or-else:
It would also be nice to have a sample script like https://github.com/tinode/chat/blob/master/tn-cli/sample-script.txt but with macros.

aforge:
Made `macro` import optional and introduced a sample script.

tn-cli/sample-macro-script.txt
```diff
@@ -0,0 +1,46 @@
...
+
+# Create user 'Test User'
+.must $user useradd --name 'Test User' --password test123 --comment test0 \
+  --cred email:test@example.com --avatar ./test-128.jpg --tags test,test-user \
+  --auth=JRWPA --anon=JW test
+
+# Login as xena.
+.must $xena login --scheme=basic --secret=xena:xena123
+
+# Change test's publis name.
```
or-else: `s/publis/public`

tn-cli/sample-macro-script.txt
```diff
@@ -0,0 +1,46 @@
+# This script shows what can be done with tn-cli macros. Run it as
+#
+#   python tn-cli.py --no-login < sample-macro-script.txt
+#
+# The script performs the following:
+#
+#  - Logs in as Xena (user has root privileges)
+#  - Creates a new user Test User with useradd
```
or-else: First creates Test then logs in as xena.

tn-cli/sample-macro-script.txt
```diff
@@ -0,0 +1,46 @@
...
+#  - Changes Test User's name to 'Test User 1'
+#  - Modifies Test User's default anon acs to JR with chacs
+#  - Sets Test User's password to test456
+#  - Deletes Test User with userdel
+#  - Uses resolve to locate user Alice
+#  - Subscribes to Alice
+#  - Sends a message to Alice
+
+# Create user 'Test User'
```
or-else: Maybe add an explanation that `.must` stops execution if command fails and and `$user` is a variable which will hold the result of the execution.

tn-cli/tn-cli.py
```diff
@@ -10,6 +10,7 @@
 import grpc
 import json
 from PIL import Image
+import imp
```
or-else: I don't think it's currently being used. When I suggested `imp` I meant something like a command line parameter for optionally loading a macros file.
```
./tn-cli.py --load-macro=./path/to/macro.py
```

tn-cli/sample-macro-script.txt
aforge: Done.

tn-cli/sample-macro-script.txt
aforge: Done.

tn-cli/tn-cli.py
aforge: Ah, makes sense. Thanks. Added the flag.

PR #362, opened 2020-02-16
Add Chcred and VCard macros in tn-cli.

aforge:
Chcred adds or removes credentials for a user.
VCard prints user's vcard (description or credentials).


tn-cli/tn-cli.py
```diff
@@ -1044,6 +1072,7 @@ def print_server_params(params):
                 printerr("Failed to read authentication cookie", err)
 
     # Attempt to load the macro file if available.
-    macros = imp.load_source('macros', args.load_macros) if args.load_macros else None
+    import importlib
```
or-else: Maybe avoid importing the module if it's not used?

```python
macros = None
if args.load_macros:
   import importlib
   macros = importlib.import_module('macros', args.load_macros)
```

or-else: Also, `macros` is a global, right? Maybe add it to globals?

aforge: Added conditional import.
`macros` is global but it's not written anywhere inside the functions so python will know it is global.
And this specific piece of code (like 1076) is the top-level script so all variables are treaded as global.

PR #364, opened 2020-02-18
Add login to Validator.ResetSecret signature.

aforge:
Login is included in the password reset email message.

### DIFF
--- a/server/auth/anon/auth_anon.go
+++ b/server/auth/anon/auth_anon.go
@@ -58,6 +58,12 @@ func (authenticator) RestrictedTags() ([]string, error) {
 	return nil, nil
 }
 
+// GetResetParams returns authenticator parameters passed to password reset handler
+// (none for anonymous).
+func (authenticator) GetResetParams(uid types.Uid, authScheme string) (map[string]interface{}, error) {
+	return nil, nil
+}
+
 func init() {
 	store.RegisterAuthScheme("anonymous", &authenticator{})
 }
--- a/server/auth/anon/auth_anon.go
+++ b/server/auth/anon/auth_anon.go
@@ -60,7 +60,7 @@ func (authenticator) RestrictedTags() ([]string, error) {
 
 // GetResetParams returns authenticator parameters passed to password reset handler
 // (none for anonymous).
-func (authenticator) GetResetParams(uid types.Uid, authScheme string) (map[string]interface{}, error) {
+func (authenticator) GetResetParams(uid types.Uid) (map[string]interface{}, error) {
 	return nil, nil
 }
 
--- a/server/auth/auth.go
+++ b/server/auth/auth.go
@@ -236,7 +236,7 @@ type AuthHandler interface {
 	RestrictedTags() ([]string, error)
 
   // GetResetParams returns authenticator parameters passed to password reset handler
-  // for the provided user id and authentication scheme.
+  // for the provided user id.
   // Returns: map of params.
-	GetResetParams(uid types.Uid, authScheme string) (map[string]interface{}, error)
+	GetResetParams(uid types.Uid) (map[string]interface{}, error)
 }
--- a/server/auth/auth.go
+++ b/server/auth/auth.go
@@ -234,4 +234,9 @@ type AuthHandler interface {
 
 	// RestrictedTags returns the tag namespaces which are restricted by this authenticator.
 	RestrictedTags() ([]string, error)
+
+  // GetResetParams returns authenticator parameters passed to password reset handler
+  // for the provided user id and authentication scheme.
+  // Returns: map of params.
+	GetResetParams(uid types.Uid, authScheme string) (map[string]interface{}, error)
 }
--- a/server/auth/basic/auth_basic.go
+++ b/server/auth/basic/auth_basic.go
@@ -275,8 +275,8 @@ func (a *authenticator) RestrictedTags() ([]string, error) {
 }
 
 // GetResetParams returns authenticator parameters passed to password reset handler.
-func (a *authenticator) GetResetParams(uid types.Uid, authScheme string) (map[string]interface{}, error) {
-	login, _, _, _, err := store.Users.GetAuthRecord(uid, authScheme)
+func (a *authenticator) GetResetParams(uid types.Uid) (map[string]interface{}, error) {
+	login, _, _, _, err := store.Users.GetAuthRecord(uid, a.name)
 	if err != nil {
 		return nil, err
 	}
--- a/server/auth/basic/auth_basic.go
+++ b/server/auth/basic/auth_basic.go
@@ -274,6 +274,22 @@ func (a *authenticator) RestrictedTags() ([]string, error) {
 	return tags, nil
 }
 
+// GetResetParams returns authenticator parameters passed to password reset handler.
+func (a *authenticator) GetResetParams(uid types.Uid, authScheme string) (map[string]interface{}, error) {
+	login, _, _, _, err := store.Users.GetAuthRecord(uid, authScheme)
+	if err != nil {
+		return nil, err
+	}
+	// User does not have a record matching the authentication scheme.
+	if login == "" {
+		return nil, types.ErrNotFound
+	}
+
+	params := make(map[string]interface{})
+	params["login"] = login
+	return params, nil
+}
+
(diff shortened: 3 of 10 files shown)


server/validate/email/validate.go
```diff
@@ -215,14 +215,15 @@ func (v *validator) Request(user t.Uid, email, lang, resp string, tmpToken []byt
 }
 
 // ResetSecret sends a message with instructions for resetting an authentication secret.
-func (v *validator) ResetSecret(email, scheme, lang string, tmpToken []byte) error {
+func (v *validator) ResetSecret(email, scheme, lang, login string, tmpToken []byte) error {
```
or-else: This breaks the abstraction. Let me think about it a bit.

or-else:
Here is my thinking.

`Authenticator` interface uses a `secret` to authenticate a session. In case of `basic` auth scheme the secret just happens to be a login - password pair but a generic Authenticator interface does not know anything about this division. Validator interface is likewise not aware of any internal structure of the `secret`. I think it's a good idea to keep it this way.

On the other hand, the need to recover `login` is real. I think the best way to do it is for the authenticator to expose internal structure when needed. I would suggest to add a method like `GetResetParams(uid t.Uid) map[string] interface{}` to Authenticator. Then either call it in `Session.authSecretReset` and pass result to `validator.ResetSecret`:

```go
params := hdl.GetResetParams(uid)
validator.ResetSecret(credValue, authScheme, s.lang, token, params)
```
or maybe pass `hdl` itself to ResetSecret and let `validator.ResetSecret` call `hdl.GetResetParams`.
```go
validator.ResetSecret(credValue, hdl, s.lang, token)
```

Validator would add returned parameters to email template in a generic fashion.

or-else:
It's also possible to make the template specific to auth method.

aforge:
Updated the PR. Please, take a look.

brettoryan:
How do I do that mate ?

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

server/auth/basic/auth_basic.go
```diff
@@ -274,6 +274,22 @@ func (a *authenticator) RestrictedTags() ([]string, error) {
 	return tags, nil
 }
 
+// GetResetParams returns authenticator parameters passed to password reset handler.
+func (a *authenticator) GetResetParams(uid types.Uid, authScheme string) (map[string]interface{}, error) {
```
or-else: Likewise redundant `authScheme`

or-else: `a.name` is the name of authenticator.

server/auth/rest/auth_rest.go
```diff
@@ -239,6 +239,12 @@ func (a *authenticator) RestrictedTags() ([]string, error) {
 	return resp.StrSliceVal, nil
 }
 
+// GetResetParams returns authenticator parameters passed to password reset handler
+// (none for rest).
+func (authenticator) GetResetParams(uid types.Uid, authScheme string) (map[string]interface{}, error) {
```
or-else: ditto `authScheme`

or-else: It would be nice to pass request to the server here instead of returning nil, but it can be done later.

server/auth/token/auth_token.go
```diff
@@ -172,6 +172,12 @@ func (authenticator) RestrictedTags() ([]string, error) {
 	return nil, nil
 }
 
+// GetResetParams returns authenticator parameters passed to password reset handler
+// (none for token).
+func (authenticator) GetResetParams(uid types.Uid, authScheme string) (map[string]interface{}, error) {
```
or-else: ditto

server/validate/email/validate.go
```diff
@@ -215,13 +215,17 @@ func (v *validator) Request(user t.Uid, email, lang, resp string, tmpToken []byt
...
 // ResetSecret sends a message with instructions for resetting an authentication secret.
-func (v *validator) ResetSecret(email, scheme, lang, login string, tmpToken []byte) error {
+func (v *validator) ResetSecret(email, scheme, lang string, tmpToken []byte, params map[string]interface{}) error {
 	// Normalize email to make sure Unicode case collisions don't lead to security problems.
 	email = strings.ToLower(email)
 
 	token := make([]byte, base64.URLEncoding.EncodedLen(len(tmpToken)))
 	base64.URLEncoding.Encode(token, tmpToken)
 	body := new(bytes.Buffer)
+	login, ok := params["login"].(string)
```
or-else: Check if `params` is `nil`.

server/auth/anon/auth_anon.go
```diff
@@ -58,6 +58,12 @@ func (authenticator) RestrictedTags() ([]string, error) {
 	return nil, nil
 }
 
+// GetResetParams returns authenticator parameters passed to password reset handler
+// (none for anonymous).
+func (authenticator) GetResetParams(uid types.Uid, authScheme string) (map[string]interface{}, error) {
```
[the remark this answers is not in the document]
or-else: `a.name`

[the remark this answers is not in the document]
aforge: Thanks. Done.

server/auth/auth.go
```diff
@@ -234,4 +234,9 @@ type AuthHandler interface {
 
 	// RestrictedTags returns the tag namespaces which are restricted by this authenticator.
 	RestrictedTags() ([]string, error)
+
+  // GetResetParams returns authenticator parameters passed to password reset handler
+  // for the provided user id and authentication scheme.
+  // Returns: map of params.
+	GetResetParams(uid types.Uid, authScheme string) (map[string]interface{}, error)
```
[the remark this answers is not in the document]
aforge: Done.

server/auth/basic/auth_basic.go
aforge: Done.

server/auth/rest/auth_rest.go
aforge: Added a TODO.

server/auth/token/auth_token.go
aforge: Done.

server/validate/email/validate.go
aforge: done.

server/validate/email/validate.go
```diff
@@ -215,14 +215,25 @@ func (v *validator) Request(user t.Uid, email, lang, resp string, tmpToken []byt
...
 // ResetSecret sends a message with instructions for resetting an authentication secret.
-func (v *validator) ResetSecret(email, scheme, lang string, tmpToken []byte) error {
+func (v *validator) ResetSecret(email, scheme, lang string, tmpToken []byte, params map[string]interface{}) error {
 	// Normalize email to make sure Unicode case collisions don't lead to security problems.
 	email = strings.ToLower(email)
 
 	token := make([]byte, base64.URLEncoding.EncodedLen(len(tmpToken)))
 	base64.URLEncoding.Encode(token, tmpToken)
 	body := new(bytes.Buffer)
+	var login string
```
or-else: `login := "-"` would simplify logic

or-else: Even better would be to leave login empty if missing an then use template condition to include login only if it's not missing, like
```
{{if .Login}} 
Login: {{.Login}} 
{{end}}
```

or-else: even better:
```
{{with .Login}} 
Login: {{.}} 
{{end}}
```

aforge: Thanks! Done.

server/validate/email/validate.go
```diff
@@ -224,13 +224,7 @@ func (v *validator) ResetSecret(email, scheme, lang string, tmpToken []byte, par
...
 	var login string
 	if params != nil {
-		var ok bool
-		login, ok = params["login"].(string)
-		if !ok || login == "" {
-			login = "-"
-		}
-	} else {
-		login = "-"
+		login = params["login"].(string)
```
or-else: Maybe it's a good idea to panic if `params["login"]` is not a string. If so, please add a comment to that extent. If not add a `_`:`login, _ = params["login"].(string)`

aforge: Added a comment.

PR #372, opened 2020-03-03
A couple bugfixes in exporter.

aforge:


monitoring/exporter/main.go
```diff
@@ -150,6 +156,6 @@ func main() {
 	}
 
 	log.Println("Reading Tinode expvar from", *tinodeAddr)
-	log.Printf("Serving metrics at %s. Server type %s", *listenAt, *serveFor)
+	log.Printf("Serving metrics at %s. Server type %s", *listenAt, serverTypeString)
```
or-else: It's not clear what is being served at `listenAt` in case of influxdb.
Also maybe add a line like "Writing metrics to ...".

aforge: Changed to "Exporter running at".

PR #376, opened 2020-03-07
Memory leak in cluster sessions, cosmetics for monitoring

or-else:
Changes:
1. Memory leak, line 765 in `server/cluster.go`
https://github.com/tinode/chat/compare/devel...cluster2?expand=1#diff-d412b9240514b30a308e4b4c8208ae72R765
2. Removed initialization of `Session.remoteSubs` from proxied cluster sessions because it's not used.
3. Added check for duplicate session IDs.
4. Some cosmetic changes for monitoring: version number as decimal as opposite to hex, uptime.
5. Some debug logging removed, some `Printf`s replaced with `Println`s as the latter is better performing. 
6. fixed a small memory waste in long polling. It's minor and unrelated to that massive leak.

### DIFF
--- a/server/cluster.go
+++ b/server/cluster.go
@@ -183,7 +183,7 @@ func (n *ClusterNode) reconnect() {
 			n.reconnecting = false
 			n.lock.Unlock()
 			statsInc("LiveClusterNodes", 1)
-			log.Printf("cluster: connection to '%s' established", n.name)
+			log.Println("cluster: connected to", n.name)
 			return
 		} else if count == 0 {
 			reconnTicker = time.NewTicker(defaultClusterReconnect)
@@ -196,7 +196,7 @@ func (n *ClusterNode) reconnect() {
 			// Wait for timer to try to reconnect again. Do nothing if the timer is inactive.
 		case <-n.done:
 			// Shutting down
-			log.Printf("cluster: node '%s' shutdown started", n.name)
+			log.Println("cluster: shutdown started at node", n.name)
 			reconnTicker.Stop()
 			if n.endpoint != nil {
 				n.endpoint.Close()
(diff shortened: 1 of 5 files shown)


server/cluster.go
```diff
@@ -752,14 +760,19 @@ func (sess *Session) rpcWriteLoop() {
 
 		case topic := <-sess.detach:
 			sess.delSub(topic)
+
+			// All proxied subsriptions are gone, this session is no longer needed.
+			if sess.countSub() == 0 {
```
aforge: Nice find!

server/sessionstore.go
```diff
@@ -63,41 +72,41 @@ func (ss *SessionStore) NewSession(conn interface{}, sid string) (*Session, int)
...
-			if sess.lastTouched.Before(expire) {
-				ss.lru.Remove(elem)
-				delete(ss.sessCache, sess.sid)
-				expired = append(expired, sess)
-			} else {
-				break // don't need to traverse further
-			}
+	ss.sessCache[s.sid] = &s
+
+	// Expire stale long polling sessions.
```
aforge: This is confusing from just reading the code.
One needs to remember that `SessionStore.lru` is only used for `LPOLL`. I suggest either adding a comment explaining it or moving this code back to the s.proto == LPOLL block above.

or-else: Expanded the comment. 

The `s.proto == LPOLL` was what caused the problem. It limited cleaning up of stale LP sessions only to when a new LP session was initialized.

server/cluster.go
or-else: :)

server/cluster.go
```diff
@@ -279,7 +279,7 @@ func (n *ClusterNode) callAsync(proc string, msg, resp interface{}, done chan *r
 
 // Proxy forwards message to master
 func (n *ClusterNode) forward(msg *ClusterReq) error {
-	log.Printf("cluster: forwarding request to node '%s'", n.name)
+	// log.Printf("cluster: forwarding request to node '%s'", n.name)
```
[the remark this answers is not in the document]
or-else: Done

PR #377, opened 2020-03-09
Allow bundling all available DB adapters in tinode binaries.

aforge:
In addition to specifying the DB adapter explicitly via build tags,
allow bundling all available DB adapters to the binary (enabled via "all" tag).
The adapter may be chosen by setting `store_config.adapter_name` field in `tinode.conf`.

Modify build scripts and docker configs to support this.
Docker tinode-all image includes the bundled binaries, the adapter may be
chosen by providing ADAPTER_NAME env var.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -48,8 +48,8 @@ See [instructions](./docker/README.md)
 	```
  - **All** (bundle all the above DB adapters):
 	```
-	go get -tags all github.com/tinode/chat/server && go build -tags all -o $GOPATH/bin/tinode github.com/tinode/chat/server
-	go get -tags all github.com/tinode/chat/tinode-db && go build -tags all -o $GOPATH/bin/init-db github.com/tinode/chat/tinode-db
+	go get -tags "mysql rethinkdb mongodb" github.com/tinode/chat/server && go build -tags "mysql rethinkdb mongodb" -o $GOPATH/bin/tinode github.com/tinode/chat/server
+	go get -tags "mysql rethinkdb mongodb" github.com/tinode/chat/tinode-db && go build -tags "mysql rethinkdb mongodb" -o $GOPATH/bin/init-db github.com/tinode/chat/tinode-db
 	```
 
 	Note the required **`-tags rethinkdb`**, **`-tags mysql`**, **`-tags mongodb`** or **`-tags all`** build option.
@@ -73,7 +73,7 @@ See [instructions](./docker/README.md)
 ```js
 	"store_config: {
 		...
-		"adapter_name": "mysql",
+		"use_adapter": "mysql",
 		...
 	},
 ```
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -52,7 +52,7 @@ See [instructions](./docker/README.md)
 	go get -tags "mysql rethinkdb mongodb" github.com/tinode/chat/tinode-db && go build -tags "mysql rethinkdb mongodb" -o $GOPATH/bin/init-db github.com/tinode/chat/tinode-db
 	```
 
-	Note the required **`-tags rethinkdb`**, **`-tags mysql`**, **`-tags mongodb`** or **`-tags all`** build option.
+	Note the required **`-tags rethinkdb`**, **`-tags mysql`** or **`-tags mongodb`** build option.
 
 	You may also optionally define `main.buildstamp` for the server by adding a build option, for instance, with a timestamp:
 	```
@@ -69,7 +69,7 @@ See [instructions](./docker/README.md)
 	},
 ```
 
-5. If you bundling all available DB adapters in the Tinode binary, make sure you specify the adapter name in your `tinode.conf`. E.g. you want to run Tinode with MySQL:
+5. Make sure you specify the adapter name in your `tinode.conf`. E.g. you want to run Tinode with MySQL:
 ```js
 	"store_config: {
 		...
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -46,8 +46,13 @@ See [instructions](./docker/README.md)
 	go get -tags mongodb github.com/tinode/chat/server && go build -tags mongodb -o $GOPATH/bin/tinode github.com/tinode/chat/server
 	go get -tags mongodb github.com/tinode/chat/tinode-db && go build -tags mongodb -o $GOPATH/bin/init-db github.com/tinode/chat/tinode-db
 	```
+ - **All** (bundle all the above DB adapters):
+	```
+	go get -tags all github.com/tinode/chat/server && go build -tags all -o $GOPATH/bin/tinode github.com/tinode/chat/server
+	go get -tags all github.com/tinode/chat/tinode-db && go build -tags all -o $GOPATH/bin/init-db github.com/tinode/chat/tinode-db
+	```
 
-	Note the required **`-tags rethinkdb`**, **`-tags mysql`** or **`-tags mongodb`** build option.
+	Note the required **`-tags rethinkdb`**, **`-tags mysql`**, **`-tags mongodb`** or **`-tags all`** build option.
 
 	You may also optionally define `main.buildstamp` for the server by adding a build option, for instance, with a timestamp:
 	```
@@ -64,7 +69,16 @@ See [instructions](./docker/README.md)
 	},
 ```
 
-5. Now that you have built the binaries, follow instructions in the _Running a Standalone Server_ section.
+5. If you bundling all available DB adapters in the Tinode binary, make sure you specify the adapter name in your `tinode.conf`. E.g. you want to run Tinode with MySQL:
+```js
+	"store_config: {
+		...
+		"adapter_name": "mysql",
+		...
+	},
+```
+
+6. Now that you have built the binaries, follow instructions in the _Running a Standalone Server_ section.
 
 ## Running a Standalone Server
 
--- a/build-all.sh
+++ b/build-all.sh
@@ -5,7 +5,6 @@
 # If directory ./server/static exists, it's asumed to contain
 # TinodeWeb and then it's also copied and archived.
 
-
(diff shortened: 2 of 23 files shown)


INSTALL.md
```diff
@@ -64,7 +69,16 @@ See [instructions](./docker/README.md)
 	},
 ```
 
-5. Now that you have built the binaries, follow instructions in the _Running a Standalone Server_ section.
+5. If you bundling all available DB adapters in the Tinode binary, make sure you specify the adapter name in your `tinode.conf`. E.g. you want to run Tinode with MySQL:
+```js
+	"store_config: {
+		...
+		"adapter_name": "mysql",
```
or-else: I would use `use_adapter` key for consistency. For example:
```
	// Large media/blob handlers.
	"media": {
		// Media handler to use
		"use_handler": "fs",
```

INSTALL.md
```diff
@@ -46,8 +46,13 @@ See [instructions](./docker/README.md)
 	go get -tags mongodb github.com/tinode/chat/server && go build -tags mongodb -o $GOPATH/bin/tinode github.com/tinode/chat/server
 	go get -tags mongodb github.com/tinode/chat/tinode-db && go build -tags mongodb -o $GOPATH/bin/init-db github.com/tinode/chat/tinode-db
 	```
+ - **All** (bundle all the above DB adapters):
```
or-else: A couple of points:

1. A new tag is not really needed. We just need to add support for multiple tags like `go build -tags="mysql rethinkdb mongodb"`. This is more flexible than "all" and requires fewer code changes.

2. I think `all` is too generic. It's not namespaced to databases. It's just `all` for the entire project. Maybe `all_databases` and such.

docker/README.md
```diff
@@ -53,6 +53,11 @@ All images are available at https://hub.docker.com/r/tinode/
 	$ docker run -p 6060:18080 -d --name tinode-srv --network tinode-net tinode/tinode-mongodb:latest
 	```
 
+	You can also run Tinode with the `tinode/tinode-all` image (which has all of the above DB adapters compiled in). You will need to specify the database adapter using the `ADAPTER_NAME` environment variable. E.g. for `mysql`, the command line will look like
```
or-else: Maybe name it `tinode/tinode` like everyone else does.

docker/README.md
```diff
@@ -61,6 +66,7 @@ All images are available at https://hub.docker.com/r/tinode/
 	 * [MySQL tags](https://hub.docker.com/r/tinode/tinode-mysql/tags/)
 	 * [RethinkDB tags](https://hub.docker.com/r/tinode/tinode-rethink/tags/)
 	 * [MongoDB tags](https://hub.docker.com/r/tinode/tinode-mongodb/tags/) (comming soon)
```
or-else: It's already available

docker/README.md
```diff
@@ -127,6 +133,7 @@ You can specify the following environment variables when issuing `docker run` co
 
 | Variable | Type | Default | Function |
 | --- | --- | --- | --- |
+| `ADAPTER_NAME` | string |  | DB adapter name (use with `tinode/tinode-all` container only) |
```
or-else: Maybe `STORE_USE_ADAPTER`? Namespacing to store more consistent with `MEDIA_HANDLER` (which is less than ideal itself, it should be named `MEDIA_USE_HANDLER`).

server/store/store.go
```diff
@@ -37,7 +43,22 @@ func openAdapter(workerId int, jsonconf json.RawMessage) error {
...
+				adp = ad
+			} else {
+				return errors.New("store: " + config.AdapterName + " adapter is not available in this binary")
+			}
+		} else if len(availableAdapters) == 1 {
+			// Attempt to default to the only entry in availableAdapters.
+			for _, v := range availableAdapters {
+				adp = v
+			}
+			log.Println("store: db adapter not specified explicitly; using", adp.GetName())
```
or-else: I'm not sure this is a good idea. Let it fail if misconfigured.

docker/README.md
or-else: BTW, it should not be a problem if it's used with any other docker.

docker/README.md
aforge: Done.

docker/README.md
aforge: Done.

docker/README.md
aforge: Switched to `STORE_USE_ADAPTER`.

INSTALL.md
```diff
@@ -46,8 +46,13 @@ See [instructions](./docker/README.md)
...
+	```
+	go get -tags "mysql rethinkdb mongodb" github.com/tinode/chat/server && go build -tags "mysql rethinkdb mongodb" -o $GOPATH/bin/tinode github.com/tinode/chat/server
+	go get -tags "mysql rethinkdb mongodb" github.com/tinode/chat/tinode-db && go build -tags "mysql rethinkdb mongodb" -o $GOPATH/bin/init-db github.com/tinode/chat/tinode-db
+	```
 
-	Note the required **`-tags rethinkdb`**, **`-tags mysql`** or **`-tags mongodb`** build option.
+	Note the required **`-tags rethinkdb`**, **`-tags mysql`**, **`-tags mongodb`** or **`-tags all`** build option.
```
or-else: `-tags all` -> `-tags alldbs`.

INSTALL.md
```diff
@@ -64,7 +69,16 @@ See [instructions](./docker/README.md)
 	},
 ```
 
-5. Now that you have built the binaries, follow instructions in the _Running a Standalone Server_ section.
+5. If you bundling all available DB adapters in the Tinode binary, make sure you specify the adapter name in your `tinode.conf`. E.g. you want to run Tinode with MySQL:
```
or-else: `all available DB adapters` -> `more than one DB adapter`

build-all.sh
```diff
@@ -16,7 +15,8 @@ goplat=( darwin windows linux )
 # Supported CPU architectures: amd64
 goarc=( amd64 )
 # Supported database tags
-dbtags=( mysql mongodb rethinkdb )
+dbadapters=( mysql mongodb rethinkdb )
+dbtags=( ${dbadapters[@]} alldbs )
```
or-else: What's the value of splitting it into two vars, why not like you did before `( mysql mongodb rethinkdb alldbs )`?

docker-release.sh
```diff
@@ -20,7 +31,7 @@ if [[ ${ver[2]} != *"-"* ]]; then
   FULLRELEASE=1
 fi
 
-dbtags=( mysql mongodb rethinkdb )
+dbtags=( mysql mongodb rethinkdb all )
```
or-else: `all` -> `alldbs`

docker-release.sh
```diff
@@ -31,15 +42,16 @@ docker login -u $user -p $pass
 # Remove earlier builds
 for dbtag in "${dbtags[@]}"
 do
+  name="$(containerName $dbtag)"
```
or-else: I think this is going to fail: 
1. Space in the name
2. `tinode alldbs` won't exists because you names it just `tinode`

docker-release.sh
```diff
@@ -54,12 +66,13 @@ curl -u $user:$pass -i -X DELETE \
 # Deploy images for various DB backends
 for dbtag in "${dbtags[@]}"
 do
+  name="$(containerName $dbtag)"
```
or-else: likewise, it's going to fail because of space and because `tinode alldbs` won't exist.

server/store/store.go
```diff
@@ -16,6 +16,7 @@ import (
 )
 
 var adp adapter.Adapter
+var availableAdapters map[string]adapter.Adapter = make(map[string]adapter.Adapter)
```
or-else: `var availableAdapters map[string]adapter.Adapter = ...` -> `availableAdapters := ...` or 
`var availableAdapters  = ...`

server/store/store.go
```diff
@@ -37,7 +40,16 @@ func openAdapter(workerId int, jsonconf json.RawMessage) error {
 	}
 
 	if adp == nil {
-		return errors.New("store: database adapter is missing")
+		if len(config.UseAdapter) > 0 {
+			// Adapter name specified explicitly.
```
or-else: Maybe also for compatibility with the previous version check if `len(config.UseAdapter) == 1` and then just use it.

INSTALL.md
aforge: Done.

INSTALL.md
aforge: Changed the wording. `use_adapter` must always be present in `tinode.conf`.

build-all.sh
aforge: I did it so it's easy to concatenate all available db adapters.

docker-release.sh
aforge: Thanks, good caatch.

docker-release.sh
aforge: No, `containerName` is a function.
`$(containerName $dbtag)` will evaluate to `tinode`.

server/store/store.go
aforge: Done.

server/store/store.go
aforge: Wait, but `len(config.UseAdapter)` returns the length of the string.
If you mean `len(config.Adapters) == 1` and using the only available adapter,
isn't that something we've discussed: https://github.com/tinode/chat/pull/377#discussion_r389472121

or-else: Yes, I meant `len(config.Adapters) == 1`.

Apparently I misunderstood your code change. Sorry. I think automatically selecting the adapter when `len(config.Adapters) == 1` is a good thing. Otherwise we would force everyone to update their conf files. And the log entry is not needed. We are printing selected adapter in `main.go`.

aforge: Done.

PR #381, opened 2020-03-15
fix for double routing of {sub topic=new} in cluster

or-else:
The other option was to generate topic name on the session origin node then send it to the master node which is a bit more complex. I think this approach is ok while the number of nodes is not too large (~20 nodes should be fine).

### DIFF
--- a/server/cluster.go
+++ b/server/cluster.go
@@ -323,7 +323,7 @@ type Cluster struct {
 
 // Master at topic's master node receives C2S messages from topic's proxy nodes.
 // The message is treated like it came from a session: find or create a session locally,
-// dispatch the message to it like it came from a normal ws/lp connection.
+// dispatch the message to it like it came from a normal ws/lp/gRPC connection.
 // Called by a remote node.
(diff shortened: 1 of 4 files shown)


server/cluster.go
```diff
@@ -523,6 +523,21 @@ func (c *Cluster) isRemoteTopic(topic string) bool {
...
+// genLocalTopicName is just like genTopicName(), but the generated name belongs to the current cluster node.
+func (c *Cluster) genLocalTopicName() string {
+	topic := genTopicName()
+	if c == nil {
+		// Cluster not initialized, all topics are local
+		return topic
+	}
+
+	// FIXME: if cluster is large it may become too inefficient.
+	for c.ring.Get(topic) != c.thisNodeName {
```
aforge: Humm... would it be possible to generate the name on the node where the original session sits? This way, we'll land on the correct node from the very start.

or-else: This is exactly that. It generates topic name on the origin node making sure the new topic is hosted on the current node.

PR #382, opened 2020-03-17
TNPG: send pushes in batches.

aforge:


server/push/tnpg/push_tnpg.go
```diff
@@ -26,6 +26,7 @@ type configType struct {
 	Enabled          bool `json:"enabled"`
 	Buffer           int  `json:"buffer"`
 	CompressPayloads bool `json:"compress_payloads"`
+	BatchSize        int  `json:"batch_size"`
```
or-else: FCM SendAll supports batches of 100 messages and that's it:
https://godoc.org/firebase.google.com/go/messaging#Client.SendAll

Configuring more than 100 won't work. Configuring fewer is just wasteful. I see no point in letting users configure this value if FCM does not let us configure it. A hardcoded 100 seems like the best choice.

aforge: Sure.

server/push/tnpg/push_tnpg.go
```diff
@@ -23,10 +23,11 @@ type Handler struct {
...
 
 type configType struct {
-	Enabled          bool `json:"enabled"`
-	Buffer           int  `json:"buffer"`
-	CompressPayloads bool `json:"compress_payloads"`
-	BatchSize        int  `json:"batch_size"`
+	Enabled          bool   `json:"enabled"`
+	Buffer           int    `json:"buffer"`
+	CompressPayloads bool   `json:"compress_payloads"`
+	BatchSize        int    `json:"batch_size"`
```
or-else: As I wrote in the other PR I would make this hardcoded 100.

aforge: This is an outdated snapshot. I removed it.

PR #387, opened 2020-03-20
Tinode docker consolidation.

aforge:
As a part of preparing docker-compose setup.

### DIFF
--- a/docker/tinode/Dockerfile
+++ b/docker/tinode/Dockerfile
@@ -24,10 +24,14 @@ LABEL version=$VERSION
 # Alternatively use
 # `--build-arg TARGET_DB=mysql` to build for MySQL or
 # `--build-arg TARGET_DB=mongodb` to build for MongoDB.
+# `--build-arg TARGET_DB=alldbs` to build a generic Tinode docker image.
 ARG TARGET_DB=rethinkdb
 ENV TARGET_DB=$TARGET_DB
 
 # Runtime options.
+
+# Specifies what jobs to run: init-db, tinode or both.
+ENV SERVICES_TO_RUN='both'
 
 # An option to reset database.
 ENV RESET_DB=false
--- a/docker/tinode/entrypoint.sh
+++ b/docker/tinode/entrypoint.sh
@@ -71,8 +71,56 @@ else
 	echo "" > $STATIC_DIR/firebase-init.js
 fi
 
-# Initialize the database if it has not been initialized yet or if data reset/upgrade has been requested.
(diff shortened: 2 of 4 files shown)


docker/tinode/Dockerfile
```diff
@@ -24,11 +24,15 @@ LABEL version=$VERSION
...
 # `--build-arg TARGET_DB=mysql` to build for MySQL or
 # `--build-arg TARGET_DB=mongodb` to build for MongoDB.
+# `--build-arg TARGET_DB=alldbs` to build a generic Tinode docker image.
 ARG TARGET_DB=rethinkdb
 ENV TARGET_DB=$TARGET_DB
 
 # Runtime options.
 
+# Specifies what jobs to run: init-db, tinode or both.
+ENV SERVICES_TO_RUN='both'
```
or-else: I think this can be done better. We have the following cases to support:

1. "Trying out"/sandbox: download the container, run it, everything works, equivalent of what's running at sandbox.tinode.co: init db, load sample data, run tinode. This is probably the most common use case.
2. First production run: init db without loading sample data, run tinode.
3. Run container after stopping: just run tinode.
4. Upgrade DB: run init-db --upgrade=true, run tinode.
5. Reset DB: with or without loading sample data.

We can check if the container is started for the first time by checking for a local "cookie" file. For case 1, 2, 3 If the file is missing, run init-db then write the file. Otherwise just run tinode (3).

There is a question if we should load sample data by default. Either way is OK, but we should support three cases:
a. Load stock sample data
b. Load custom sample data
c. Do not load sample data at all

I think the default should be a. b & c can be controlled by supplying a path to sample data, including an empty path. The empty path is currently not possible. 

 The most common case should be the simplest.
 
Cases 4 & 5 can only be triggered with an explicit ENV var, so the way it's done now is good.

docker/tinode/entrypoint.sh
```diff
@@ -71,8 +71,56 @@ else
...
+  ;;
+esac
+
+echo "Will run init-db: ${run_init_db}, tinode: ${run_tinode}"
+
+touch /botdata/tino-password
+
+if [ "$run_init_db" == "true" ]; then
+  # Initialize the database if it has not been initialized yet or if data reset/upgrade has been requested.
+  ./init-db --reset=${RESET_DB} --upgrade=${UPGRADE_DB} --config=${CONFIG} | grep "usr;tino;" > /botdata/tino-password
```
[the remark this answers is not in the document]
aforge: Good catch. Added loading sample data back.

docker/tinode/Dockerfile
aforge: Per discussion, will parse and process init-db output.
Will do in a separate PR.

PR #389, opened 2020-03-20
Another attempt at fixing memory leak

or-else:
The previous change fixed the cases when the topic was deleted. It did not fix the most common use case of unsubscribing from all topics.

### DIFF
--- a/server/cluster.go
+++ b/server/cluster.go
@@ -22,6 +22,8 @@ const (
 	defaultClusterReconnect = 200 * time.Millisecond
 	// Number of replicas in ringhash
 	clusterHashReplicas = 20
+	// Period for running health check on cluster session: terminate sessions with no subscriptions.
+	clusterSessionCleanup = 5 * time.Second
 )
 
 type clusterNodeConfig struct {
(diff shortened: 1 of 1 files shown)


server/cluster.go
```diff
@@ -744,6 +746,8 @@ func (sess *Session) rpcWriteLoop() {
 		sess.unsubAll()
 	}()
 
+	heartBeat := time.NewTimer(clusterSessionCleanup)
```
aforge: Can you add a short comment here mentioning that this is needed for GCing orphaned sessions?

or-else:
Actually, this needs more work. The current fix has a potential for a data race:
1. Session has no subscriptions but GC is not yet triggered.
2. A new request comes for this session, while session is still in SessionStore, so `sess != nil`:
https://github.com/tinode/chat/blob/master/server/cluster.go#L355
3. Session is GC-ed before `sess.dispatch(msg.CliMsg)` is called:
https://github.com/tinode/chat/blob/master/server/cluster.go#L376
4. Message `msg.CliMsg` sent to `sess.dispatch` is lost.

or-else:
One way to solve it is to store list of nodes in the origin session. We store list of remote nodes as a list of subscribed remote topics, so store all nodes where the origin has a proxy session.

or-else:
Per my private comments, let's merge it with understanding that we need to refactor it. I've added comments to problematic code.

PR #391, opened 2020-03-21
SELECT tinode db before updating in mariadb.

benabel:


or-else:
Thanks!

benabel:
Faster than light. Have a good day.

Issue #390, opened 2020-03-21
I can't Build Source project


komeil-sh:
### **This is Error**
`GOROOT=C:\Go #gosetup
GOPATH=C:\Users\KOMEIL-PC\go #gosetup
C:\Go\bin\go.exe build -tags mongodb -o C:\Users\KOMEIL-PC\AppData\Local\Temp\___go_build_main_go.exe C:/Users/KOMEIL-PC/Downloads/Compressed/chat-master/server/main.go #gosetup
..\..\..\..\go\src\github.com\tinode\chat\server\media\s3\s3.go:14:2: cannot find package "github.com/aws/aws-sdk-go/aws" in any of:
	C:\Go\src\github.com\aws\aws-sdk-go\aws (from $GOROOT)
	C:\Users\KOMEIL-PC\go\src\github.com\aws\aws-sdk-go\aws (from $GOPATH)
..\..\..\..\go\src\github.com\tinode\chat\server\media\s3\s3.go:15:2: cannot find package "github.com/aws/aws-sdk-go/aws/awserr" in any of:
	C:\Go\src\github.com\aws\aws-sdk-go\aws\awserr (from $GOROOT)
	C:\Users\KOMEIL-PC\go\src\github.com\aws\aws-sdk-go\aws\awserr (from $GOPATH)
..\..\..\..\go\src\github.com\tinode\chat\server\media\s3\s3.go:16:2: cannot find package "github.com/aws/aws-sdk-go/aws/credentials" in any of:
	C:\Go\src\github.com\aws\aws-sdk-go\aws\credentials (from $GOROOT)
	C:\Users\KOMEIL-PC\go\src\github.com\aws\aws-sdk-go\aws\credentials (from $GOPATH)
..\..\..\..\go\src\github.com\tinode\chat\server\media\s3\s3.go:17:2: cannot find package "github.com/aws/aws-sdk-go/aws/request" in any of:
	C:\Go\src\github.com\aws\aws-sdk-go\aws\request (from $GOROOT)
	C:\Users\KOMEIL-PC\go\src\github.com\aws\aws-sdk-go\aws\request (from $GOPATH)
..\..\..\..\go\src\github.com\tinode\chat\server\media\s3\s3.go:18:2: cannot find package "github.com/aws/aws-sdk-go/aws/session" in any of:
	C:\Go\src\github.com\aws\aws-sdk-go\aws\session (from $GOROOT)
[...]

Issue #383, opened 2020-03-18
build source error!


Sopage:
### Subject of the issue
$:./build-all.sh
error message:
Stderr: cloud.google.com/go/iam
../../../cloud.google.com/go/iam/iam.go:135:58: cannot use conn (type grpc.ClientConnInterface) as type *grpc.ClientConn in argument to iam.NewIAMPolicyClient: need type assertioncloud.google.com/go/firestore/apiv1
../../../cloud.google.com/go/firestore/apiv1/firestore_client.go:226:41: cannot use connPool (type "google.golang.org/api/internal".ConnPool) as type *"google.golang.org/grpc".ClientConn in argument to firestore.NewFirestoreClient

or-else:
Works for me:
```
Genes-MacBook-Pro:chat gene$ ./build-all.sh 
Releasing 0.16.4-27-g7905bf5
Building mysql-darwin/amd64...
Building mongodb-darwin/amd64...
Building rethinkdb-darwin/amd64...
Building alldbs-darwin/amd64...
Building mysql-windows/amd64...
Building mongodb-windows/amd64...
Building rethinkdb-windows/amd64...
Building alldbs-windows/amd64...
Building mysql-linux/amd64...
Building mongodb-linux/amd64...
Building rethinkdb-linux/amd64...
Building alldbs-linux/amd64...
Building the binary for web.tinode.co
Building python code...
Packaging python tinode-grpc...
Packaging chatbot.py...
Packaging tn-cli...
```

If you need help, please give s much information about your setup as possible.

or-else:
Also keep in mind that the error is generated while compiling Google's code. It's highly unlikely that Google shipped code which does not compile.

Sopage:
thanks,  get dependencies I use "go get -v. / ...", so I don't know what's error

Sopage:
https://github.com/googleapis/google-cloud-go
select tag v0.49.0
Why not use go mod?

Sopage:
Releasing 0.16.4-3-g4184c45
Building mysql-linux/amd64...
TinodeWeb not found, skipping
Building the binary for web.tinode.co
Building python code...
Packaging python tinode-grpc...
Packaging chatbot.py...
Packaging tn-cli...



!!!!!!!!!! TinodeWeb not found, skipping. !!!!!!!!!!??

or-else:
> Why not use go mod?

Because it's relatively new and we have not migrated to go mod yet. Would you like to send a pull request?

[...]

PR #392, opened 2020-03-21
Using go mod for package management

or-else:
Fix for https://github.com/tinode/chat/issues/390 and https://github.com/tinode/chat/issues/383


or-else:
There is no point in an actual code review, these files are generated by `go mod`. Just check that it builds for you without any issues. Thanks.

komeil-sh:
Solved, thank you

PR #393, opened 2020-03-22
Run tinode only if init-db returns success.

aforge:
In tinode:tinode docker:
1. Check stderr of init-db for the presence of `All done` string.
2. If the string isn't found, stop. Otherwise, proceed to running tinode.
3. Use `SAMPLE_DATA` env var to specify the sample data file.


docker/tinode/Dockerfile
```diff
@@ -102,6 +102,7 @@ RUN tar -xzf tinode-$TARGET_DB.linux-amd64.tar.gz \
 	&& rm tinode-$TARGET_DB.linux-amd64.tar.gz
 
 # Copy config template to the container.
```
or-else: This comment should be one line lower.

docker/tinode/Dockerfile
```diff
@@ -102,6 +102,7 @@ RUN tar -xzf tinode-$TARGET_DB.linux-amd64.tar.gz \
...
+COPY init-db .
```
or-else: I believe this line copies init-db from your host docker directory to container. You should already have it in your container after downloading and unpacking `tinode-$TARGET_DB.linux-amd64.tar.gz`.

docker/tinode/entrypoint.sh
```diff
@@ -50,8 +50,8 @@ else
 fi
 
 # Load default sample data when generating or resetting the database.
```
or-else: Update comment?

docker/tinode/entrypoint.sh
```diff
@@ -89,42 +89,32 @@ if [ ! -z "$IOS_UNIV_LINKS_APP_ID" ] ; then
...
+# If sample data was provided, try to find Tino password.
+if [ ! -z "$SAMPLE_DATA" ] ; then
+	grep "usr;tino;" $init_stdout > /botdata/tino-password
 fi
 
-if [ "$run_tinode" != "true" ]; then
-  # If we don't want to run tinode, we are done.
-  exit 0
+# Check if the init-db output contains the magic string.
+grep -q "All done" $init_stderr
```
or-else: Please consider checking for exit code:
```
if [ $? -eq 0 ]
then
  echo "Success"
else
  echo "Failure" >&2
fi
```

tinode-db/main.go
```diff
@@ -225,11 +225,20 @@ func main() {
 		// Upgrade DB from one version to another.
 		err = store.UpgradeDb(config.StoreConfig)
 		if err == nil {
-			log.Println("Database successfully upgraded")
+			log.Println("Database successfully upgraded. All done.")
```
or-else: Maybe `os.Exit(0)` and `os.Exit(1)` would be more idiomatic.

or-else:
> Check stderr of init-db for the presence of All done string.

If we have to modify init-db anyway, `os.Exit()` would be a better solution.

tinode-db/main.go
aforge: Do you mean checking for the exit code instead of the magic string?

docker/tinode/entrypoint.sh
aforge: Done.

docker/tinode/entrypoint.sh
aforge: Done below.

docker/tinode/Dockerfile
aforge: This line is not necessary. Removed.

docker/tinode/entrypoint.sh
or-else: Then you don't need the `grep -q "All done" $init_stderr` line anymore.

tinode-db/main.go
or-else: Yes. Sorry for not making myself clear.

aforge: Done.

docker/tinode/entrypoint.sh
```diff
@@ -49,9 +49,9 @@ else
   STATIC_DIR="./static"
 fi
 
-# Load default sample data when generating or resetting the database.
-if [[ -z "$SAMPLE_DATA" && "$UPGRADE_DB" = "false" ]] ; then
-	SAMPLE_DATA="$DEFAULT_SAMPLE_DATA"
+# Do not load data when upgrading database.
+if [[ "$UPGRADE_DB" = "true" ]] ; then
+	SAMPLE_DATA=""
```
or-else: Just `SAMPLE_DATA=`

docker/tinode/entrypoint.sh
```diff
@@ -89,42 +89,22 @@ if [ ! -z "$IOS_UNIV_LINKS_APP_ID" ] ; then
...
-  # Maybe load sample data?
-  if [ ! -z "$SAMPLE_DATA" ] ; then
-    init_args+=("--data=$SAMPLE_DATA")
-  fi
-  # Initialize the database if it has not been initialized yet or if data reset/upgrade has been requested.
-  ./init-db "${init_args[@]}" | grep "usr;tino;" > /botdata/tino-password
+init_args=("--reset=${RESET_DB}" "--upgrade=${UPGRADE_DB}" "--config=${CONFIG}")
+# Maybe load sample data?
+if [ ! -z "$SAMPLE_DATA" ] ; then
+	init_args+=("--data=$SAMPLE_DATA")
```
or-else: I don't think this check is needed. Passing `--data=` should work just as well.

docker/tinode/entrypoint.sh
```diff
@@ -140,10 +120,10 @@ args=("--config=${CONFIG}" "--static_data=$STATIC_DIR")
 
 # Maybe set node name in the cluster.
 if [ ! -z "$CLUSTER_SELF" ] ; then
-  args+=("--cluster_self=$CLUSTER_SELF")
+	args+=("--cluster_self=$CLUSTER_SELF")
```
or-else: The check is redundant. Empty parameter `--cluster_self=` is equivalent to no parameter.

docker/tinode/entrypoint.sh
```diff
@@ -140,10 +120,10 @@ args=("--config=${CONFIG}" "--static_data=$STATIC_DIR")
...
 fi
 if [ ! -z "$PPROF_URL" ] ; then
-  args+=("--pprof_url=$PPROF_URL")
+	args+=("--pprof_url=$PPROF_URL")
```
or-else: Likewise, redundant check.

docker/tinode/entrypoint.sh
```diff
@@ -140,10 +120,10 @@ args=("--config=${CONFIG}" "--static_data=$STATIC_DIR")
...
 fi
 
 # Create the log directory (/var/log/tinode-`current timestamp`).
```
or-else: I'm not sure we need to setup log rotation here. I would just write to log file and be done with it. Users may have their own logging needs and methods.

tinode-db/main.go
```diff
@@ -218,18 +218,27 @@ func main() {
 		log.Println("Database reset requested")
 	} else {
 		log.Println("Database exists, DB version is correct. All done.")
-		return
+		os.Exit(0)
```
or-else: Also add `os.Exit(0)` as the very last line of this file, after `genDb(&data)`. Github does not let me add a comment there.

aforge: Done.

docker/tinode/entrypoint.sh
aforge: Done.

docker/tinode/entrypoint.sh
aforge: Removed.

PR #394, opened 2020-03-23
Better startup order in E2E setup.

aforge:
Added $WAIT_FOR envvar to the `entrypoint.sh` scripts in Tinode and Exporter docker containers.
This way the containers will wait for their dependencies to become available.

Provide default Tinode FCM push notifications configuration.


or-else:
BTW, why do we need an explicit network config in docker-compose?

aforge:
Re specifying networking config: docker compose requires this in order to assign static IP addresses to containers.

Issue #395, opened 2020-03-25
Incorrect name when deleting avatar for another user in tn-cli


aforge:
Start Tinode locally, load sample data.
Connect as xena (root user).
```
$ python tn-cli.py --host localhost:6090 --login-basic=xena:xena123
```
Find `alice` and delete her avatar via usermod:
```
tn> resolve alice
...
<= 200 ok (fnd) 
<= usrALICE
tn> usermod usrALICE --avatar='␡' 
...
<= 200 ok (me) 
<= Default user='' 
```

Alice's public name changes to Xena Pacifist Peasant.

PR #396, opened 2020-03-25
Fetch correct user data when subscribing to ME on behalf of another user

aforge:
Intended as a fix for https://github.com/tinode/chat/issues/395


server/init_topic.go
```diff
@@ -113,7 +113,19 @@ func topicInit(t *Topic, sreg *sessionJoin, h *Hub) {
...
 
-	user, err := store.Users.Get(sreg.sess.uid)
+	uid := types.ParseUserId(t.name)
+	if uid.IsZero() {
+		// Log out the session
+		sreg.sess.uid = types.ZeroUid
+		log.Println("malformed topic name (sub me): ", t.name, sreg.sess.sid)
+		return types.ErrMalformed
+	}
+	if uid != sreg.sess.uid && sreg.sess.authLvl != auth.LevelRoot {
```
or-else: I don't understand this change. Maybe it's not needed.

Issuing a command `onBehalfOf` is only available to root user:
https://github.com/tinode/chat/blob/master/server/session.go#L277
`t.name` is assigned here:
https://github.com/tinode/chat/blob/master/server/session.go#L1030
and it matches `onBahalfOf`.

server/init_topic.go
```diff
@@ -113,7 +113,19 @@ func topicInit(t *Topic, sreg *sessionJoin, h *Hub) {
 func initTopicMe(t *Topic, sreg *sessionJoin) error {
 	t.cat = types.TopicCatMe
 
-	user, err := store.Users.Get(sreg.sess.uid)
+	uid := types.ParseUserId(t.name)
+	if uid.IsZero() {
```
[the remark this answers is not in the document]
aforge: Done.

server/init_topic.go
aforge: Ah, good. Thanks, removed.

PR #399, opened 2020-03-26
Docker

or-else:
1. Chatbot container more configurable
2. Removed `TNPG_COMPRESS_PAYLOADS` as unnecessary (always true)
3. Automatic gofmt and misc stuff.

### DIFF
--- a/docker/chatbot/Dockerfile
+++ b/docker/chatbot/Dockerfile
@@ -1,11 +1,12 @@
 # Dockerfile builds an image with a chatbot (Tino) for Tinode.
 
-FROM python:3.7-alpine
+FROM python:3.7-slim
 
 ARG VERSION=0.16
 ARG LOGIN_AS=
 ARG TINODE_HOST=tinode-srv:6061
 ENV VERSION=$VERSION
+ARG BINVERS=$VERSION
 
 LABEL maintainer="Tinode Team <info@tinode.co>"
 LABEL name="TinodeChatbot"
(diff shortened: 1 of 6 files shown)


docker/chatbot/Dockerfile
```diff
@@ -1,19 +1,26 @@
...
 LABEL maintainer="Tinode Team <info@tinode.co>"
 LABEL name="TinodeChatbot"
 LABEL version=$VERSION
 
 RUN mkdir -p /usr/src/bot
+
 WORKDIR /usr/src/bot
 
+# Volume with login cookie. Not created automatically.
+# VOLUME /botdata
```
aforge: What is this for?

docker/tinode/Dockerfile
```diff
@@ -95,9 +92,6 @@ ENV TNPG_AUTH_TOKEN=
 # Tinode Push Gateway user name.
 ENV TNPG_USER=
 
-# Enable Tinode Push Gateway request payload compression.
```
aforge: Can you remove https://github.com/tinode/chat/blob/devel/docker/tinode/config.template#L114 as well please?

docker/chatbot/Dockerfile
or-else: It's a reminder which volumes have special meaning and can be mapped to external directories. I can just leave a comment without the VOLUME part but it seems clearer this way.

docker/tinode/Dockerfile
or-else: done

PR #413, opened 2020-04-07
Basic session param override struct

aforge:
This PR is purposed to establish a general pattern to override session param.

### DIFF
--- a/server/datamodel.go
+++ b/server/datamodel.go
@@ -519,8 +519,9 @@ type MsgServerInfo struct {
 }
 
 // Session parameter overrides.
+// Used when by the remote topic masters to impersonate multiple proxied sessions.
 type sessionOverrides struct {
-	// Target session id.
+	// Proxied session id.
 	sid string
 }
 
--- a/server/datamodel.go
+++ b/server/datamodel.go
@@ -519,7 +519,7 @@ type MsgServerInfo struct {
 }
 
 // Session parameter overrides.
-// Used when by the remote topic masters to impersonate multiple proxied sessions.
+// Used by the remote topic masters to impersonate multiple proxied sessions.
 type sessionOverrides struct {
 	// Proxied session id.
 	sid string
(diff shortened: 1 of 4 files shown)


server/datamodel.go
```diff
@@ -518,6 +518,13 @@ type MsgServerInfo struct {
 	SeqId int `json:"seq,omitempty"`
 }
 
+// Session parameter overrides.
+// Used when by the remote topic masters to impersonate multiple proxied sessions.
```
or-else, line 522: Used ~when~ by

server/datamodel.go
```diff
@@ -518,6 +518,13 @@ type MsgServerInfo struct {
...
+type sessionOverrides struct {
```
or-else, line 523: The alternative would be to use the regular `Session` with `proto == PROXY` or `OVERRIDE` or something like that. `sessOverrides` requires passing it in exactly the same places where `sess` is already passed. It tells me extending `Session` is better than splitting it into two. 

The number of proxied sessions is T * C, where N is the number of topics and C is the number of cluster nodes, exactly the same as the increase in the number of topics due to clustering, which looks acceptable to me. Then we leave most parameters of the overridden session as blank so they won't take up memory.

aforge: Yes, I considered extending `Session`.
Right now, I'm planning to use one `Session` object per master topic and multiplex multiple remote sessions in this session.
So if two requests to the same topic arrive from two different sessions, they will have to share their shared proxy session on the master. So modifying the state of the shared session isn't going to work here. Or did you mean something else?

Another thing I was thinking about is a thread local variable/context. `Topic.run()` is single-threaded. We can set the context when an event is polled out of the queue and clear it after the event has been processed.

server/datamodel.go
aforge: Done.

aforge:
I'm planning to use `sessionOverrides` for more session-level params such as `authLvl` and device settings.
As for `queueOutWithOverrides`, I'll change the existing method in follow-up PRs.

PR #415, opened 2020-04-09
Topic proxy: basic routing between proxy and master.

aforge:
Join request basic handling.

### DIFF
--- a/server/cluster.go
+++ b/server/cluster.go
@@ -125,8 +125,9 @@ type ClusterReq struct {
 	RcptTo string
 	// Originating session
 	Sess *ClusterSess
-	// True if the original session has disconnected
-	SessGone bool
+	// True if either the original topic (when proxying topics) is gone or
+	// the original session (when proxying sessions) has disconnected
+	Done bool
 
 	// UNroutable components of {pres} message which have to be sent intra-cluster.
 	SrvPres ClusterPresExt
@@ -323,7 +324,7 @@ func (n *ClusterNode) forwardToTopicMaster(msg *ClusterReq) error {
 	var rejected bool
 	err := n.call("Cluster.TopicMaster", msg, &rejected)
 	if err == nil && rejected {
-		err = errors.New("cluster: master2 node out of sync")
+		err = errors.New("cluster: topic master node out of sync")
 	}
 	return err
 }
@@ -375,7 +376,7 @@ func (c *Cluster) Master(msg *ClusterReq, rejected *bool) error {
 	// Find the local session associated with the given remote session.
 	sess := globals.sessionStore.Get(msg.Sess.Sid)
 
-	if msg.SessGone {
+	if msg.Done {
 		// Original session has disconnected. Tear down the local proxied session.
 		if sess != nil {
 			sess.stop <- nil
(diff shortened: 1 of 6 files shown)


server/cluster.go
```diff
@@ -386,10 +429,81 @@ func (c *Cluster) Master(msg *ClusterReq, rejected *bool) error {
 	return nil
 }
 
+func (c *Cluster) TopicMaster(msg *ClusterReq, rejected *bool) error {
+	*rejected = false
+
+	sid := "topic-proxy-" + msg.RcptTo + "-" + msg.Node
+	sess := globals.sessionStore.Get(sid)
+	if msg.SessGone {
+		// Original session has disconnected. Tear down the local proxied session.
```
or-else, line 438: Previously it meant that the session at origin got disconnected and proxied session at master must be deleted too. What does it mean here? The comment is probably incorrect too.

server/cluster.go
```diff
@@ -386,10 +429,81 @@ func (c *Cluster) Master(msg *ClusterReq, rejected *bool) error {
...
+		log.Println("cluster TopicMaster: nil topic message", msg.RcptTo)
+		*rejected = true
+		return nil
+	}
+	node := globals.cluster.nodes[msg.Node]
+	if node == nil {
+		log.Println("cluster TopicMaster: request from an unknown node", msg.Node)
+		return nil
+	}
+	if msg.Sess == nil {
```
or-else, line 459: Do we still need to pass the entire sesson?

server/cluster.go
```diff
@@ -386,10 +429,81 @@ func (c *Cluster) Master(msg *ClusterReq, rejected *bool) error {
 	return nil
 }
 
+func (c *Cluster) TopicMaster(msg *ClusterReq, rejected *bool) error {
+	*rejected = false
+
+	sid := "topic-proxy-" + msg.RcptTo + "-" + msg.Node
```
or-else, line 435: Not that important, but `msg.RcptTo + "-" + msg.Node` is sufficiently unique.

server/cluster.go
```diff
@@ -386,10 +429,81 @@ func (c *Cluster) Master(msg *ClusterReq, rejected *bool) error {
...
+			topic:    msg.RcptTo,
+			pkt:      msg.CliMsg,
+			sess:     sess,
+			created:  msg.TopicMsg.JoinReq.Created,
+			newsub:   msg.TopicMsg.JoinReq.Newsub,
+			internal: msg.TopicMsg.JoinReq.Internal,
+			// Impersonate the original session.
+			sessOverrides: &sessionOverrides{
+				sid:    msg.Sess.Sid,
+				rcptTo: msg.RcptTo,
```
or-else, line 492: This duplicates `topic: msg.RcptTo` above. Is it necessary?

server/cluster.go
or-else: Oh, I see it's an indicator that proxied topic is gone. Maybe rename it accordingly? The comment certainly needs to be updated.

server/cluster.go
```diff
@@ -897,3 +1061,44 @@ func (c *Cluster) invalidateRemoteSubs() {
...
+	for {
+		select {
+		case msg, ok := <-sess.send:
+			if !ok || sess.clnode.endpoint == nil {
+				// channel closed
+				return
+			}
+			srvMsg := msg.(*ServerComMessage)
+			if srvMsg.sessOverrides == nil {
+				log.Println("cluster: proxy session overrides not specified - ", sess.sid)
```
or-else, line 1081: How can it happen? Only if there is a bug somewhere? If so we should panic instead.

PR #416, opened 2020-04-11
Topic Proxy: Route publish/broadcast requests to master.

aforge:


server/cluster.go
```diff
@@ -185,6 +186,18 @@ type ProxyJoin struct {
 	Internal bool
 }
 
+// ProxyBroadcast contains topic broadcast request parameters.
+type ProxyBroadcast struct {
+	// Original request message id.
+	Id        string
+	// To: topic.
+	From      string
```
or-else, line 194: The comment seems to be incorrect. `From` is supposedly UID of the sender.

server/topic.go
```diff
@@ -210,6 +210,17 @@ func (t *Topic) runProxy(hub *Hub) {
 		case msg := <-t.broadcast:
 			// Content message intended for broadcasting to recipients
 			log.Printf("t[%s] broadcast %+v", t.name, msg)
+			brdc := &ProxyTopicData{
+				BroadcastReq: &ProxyBroadcast{
+					Id:        msg.id,
+					From:      msg.from,
+					Timestamp: msg.timestamp,
+					SkipSid:   msg.skipSid,
```
or-else, line 218: I think it should be set only when `pub.echo == false`.

aforge: `Session.publish()` will take care of it:
https://github.com/tinode/chat/blob/devel/server/session.go#L532

PR #417, opened 2020-04-13
Topic Proxy: Route meta messages and broadcast responses.

aforge:
This PR adds basic functionality for routing meta messages and broadcasting responses to proxied sessions.

### DIFF
--- a/server/cluster.go
+++ b/server/cluster.go
@@ -170,7 +170,8 @@ type ProxyTopicData struct {
 	JoinReq *ProxyJoin
 	// Broadcast (publish, etc.) request.
 	BroadcastReq *ProxyBroadcast
-	// MetaReq      *ProxyMeta
+	// Meta request.
+	MetaReq      *ProxyMeta
 	// LeaveReq     *ProxyLeave
 }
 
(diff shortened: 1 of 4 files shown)


server/cluster.go
```diff
@@ -483,14 +496,14 @@ func (c *Cluster) TopicMaster(msg *ClusterReq, rejected *bool) error {
...
-		go sess.topicProxyWriteLoop()
+		go sess.topicProxyWriteLoop(msg.RcptTo)
 	}
 
 	switch {
 	case msg.TopicMsg.JoinReq != nil:
-		if msg.Sess.Uid > 0 {
-			log.Println("join setting uid = ", msg.Sess.Uid)
-			msg.CliMsg.from = msg.Sess.Uid.UserId()
+		if uid > 0 {
```
or-else, line 504: @aforge 

I just noticed this. Let's use `uid.IsZero()` instead. For consistency and because `types.Uid` should be treated as opaque code. We may have to change it for federation. It would be a lot easier to change it if it's not assumed to be an `int`.

aforge: Fixed in the followup PR.

PR #422, opened 2020-04-16
Topic Proxy: leave request handling

aforge:
Pass over session leave requests to the topic master and
clean up master & proxy topics (and their helper sessions) when there are no more attached sessions to them.

Note that this PR is an intermediate version. There's more work to be done to property handle subscribe/unsubscribe.

### DIFF
--- a/server/cluster.go
+++ b/server/cluster.go
@@ -526,7 +526,6 @@ func (c *Cluster) TopicMaster(msg *ClusterReq, rejected *bool) error {
 		// If the session is not found, create it.
 		var count int
 		sess, count = globals.sessionStore.NewSession(node, sid)
-    sess.isProxy = true
 
 		log.Println("cluster: topic proxy channel started", sid, count)
 		go sess.topicProxyWriteLoop(msg.RcptTo)
@@ -1216,7 +1215,7 @@ func (sess *Session) topicProxyWriteLoop(forTopic string) {
 			if srvMsg.sessOverrides != nil {
 				switch req := srvMsg.sessOverrides.origReq.(type) {
 				case nil:
-					log.Println("origReq is nil")
+					panic("cluster: origReq is nil in session overrides")
 				case *ProxyJoin:
 					response.ProxyResp.OrigRequestType = ProxyRequestJoin
 				case *ProxyLeave:
--- a/server/cluster.go
+++ b/server/cluster.go
@@ -174,7 +174,7 @@ type ProxyTopicData struct {
 	BroadcastReq *ProxyBroadcast
 	// Meta request.
 	MetaReq      *ProxyMeta
-	// LeaveReq     *ProxyLeave
+	LeaveReq     *ProxyLeave
 }
 
 // ProxyJoin contains topic join request parameters.
@@ -206,11 +206,20 @@ type ProxyMeta struct {
 	What int
 }
 
+type ProxyLeave struct {
+	Id string
+	Topic string
+	UserId types.Uid
+	Unsub bool
+	TerminateRemoteSession bool
+}
+
 // Proxy request types.
 const (
 	ProxyRequestJoin      = 1
-	ProxyRequestMeta      = 2
-	ProxyRequestBroadcast = 3
+	ProxyRequestLeave     = 2
+	ProxyRequestMeta      = 3
+	ProxyRequestBroadcast = 4
 )
 
 // ProxyResponse contains various parameters sent back by the topic master in response a topic proxy request.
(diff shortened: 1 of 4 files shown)


server/topic.go
```diff
@@ -2956,19 +3007,28 @@ func (t *Topic) addSession(s *Session, asUid types.Uid) bool {
...
-// Removes session record if 'asUid' matches subscribed user.
-func (t *Topic) remSession(s *Session, asUid types.Uid) *perSessionData {
-	if pssd, ok := t.sessions[s]; ok && (pssd.uid == asUid || asUid.IsZero()) {
-		// Check for deferred presence notification and cancel it if found.
-		if pssd.ref != nil {
-			t.defrNotif.Remove(pssd.ref)
+// Removes session record if 'doRemove' is true and either:
+// * 'asUid' matches subscribed user.
+// * or 's' is a proxy session for a proxy-master topic connection.
+func (t *Topic) maybeRemoveSession(s *Session, asUid types.Uid, doRemove bool) *perSessionData {
```
or-else, line 3013: The logic here is for the use case when the session authenticated with a root user A is subscribed to topic on behalf of user B. It also implements support for deferred online notifications. The `pssd` (per session data) contains user id overrides + reference to deferred notification. 

The session is removed from topic when the actually subscribed user  (A==B) leaves `pssd.uid == asUid` or session is disconnected `asUid.IsZero()` regardless of user (similar to `leave.terminateProxyConnection == true`).

If session is removed from topic then non-nil `pssd` is returned. Otherwise nil is returned.

How it applies to a session from a proxy topic:

* Proxy session at master topic should have zeros and nils in `pssd`.
* `pssd` should not be checked at the master topic (when `s.isProxy` is `true`). `pssd` is for supporting onBehalfOf. This session is always `onBehalfOf` multiple users. It should be checked at proxy topic (`t.isProxy` is `true`).
* The `t.defrNotif.Remove(pssd.ref)` should not be called at the master topic. It should be called  at the origin, at the proxy topic.
* Basically, the proxy session should be unconditionally removed when the origin terminates it. Otherwise it should not be affected.

server/topic.go
```diff
@@ -338,22 +381,28 @@ func (t *Topic) runLocal(hub *Hub) {
...
 				// User wants to leave and unsubscribe.
 				// asUid must not be Zero.
-				if err := t.replyLeaveUnsub(hub, leave.sess, asUid, leave.id, nil/*leave.sessOverrides*/); err != nil {
+				if err := t.replyLeaveUnsub(hub, leave.sess, asUid, leave.id, leave.sessOverrides); err != nil {
 					log.Println("failed to unsub", err, leave.sess.sid)
 					continue
 				}
 
-			} else if pssd := t.remSession(leave.sess, asUid); pssd != nil {
+			} else if pssd := t.maybeRemoveSession(leave.sess, asUid, /*doRemove=*/!leave.sess.isProxy || leave.terminateProxyConnection); pssd != nil {
```
or-else, line 396: See comment at declaration of `maybeRemoveSession()`.

server/topic.go
```diff
@@ -338,22 +381,28 @@ func (t *Topic) runLocal(hub *Hub) {
...
 				// Just leaving the topic without unsubscribing if user is subscribed.
 
-				pud := t.perUser[pssd.uid]
+				var uid types.Uid
+				if t.isProxy {
```
or-else, line 400: This section is probably incorrect. 

* Behavior at the [proxy topic + normal session] should match that of the [normal topic + normal session]:
`pud := t.perUser[pssd.uid]` like before.
* Behavior of [normal topic + proxy session] should be different: it should not care abut updating any of `t.perUser[uid] = pud` because proxy session multiplexes many uses. 

See comment at declaration of `maybeRemoveSession()`.

server/cluster.go
```diff
@@ -511,6 +526,7 @@ func (c *Cluster) TopicMaster(msg *ClusterReq, rejected *bool) error {
 		// If the session is not found, create it.
 		var count int
 		sess, count = globals.sessionStore.NewSession(node, sid)
+    sess.isProxy = true
```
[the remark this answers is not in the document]
aforge: Done.

or-else:
I have a suggestion for improving the review process. If you don't mind, please merge `devel` into `topic-proxy` directly instead of merging it into `tp-dev`. Otherwise my changes to `devel` are mixed with your changes to `tp-dev`.

server/topic.go
or-else: OK

PR #425, opened 2020-04-20
Invalidate proxied subs on cluster rehash.

aforge:
Also, a couple of minor fixes and changes to presence notifications handling.


server/cluster.go
```diff
@@ -1157,36 +1157,23 @@ func (c *Cluster) rehash(nodes []string) []string {
 	return ringKeys
 }
 
-// Iterates over sessions hosted on this node and for each session
+// invalidateProxySubs iterates over sessions proxied on this node and for each session
```
or-else, line 1160: It would be nice to be able to re-hook proxy topic to the new master. But that can be done later.

aforge: Fair enough. Added a TODO.

PR #428, opened 2020-04-24
Topic proxy: more bug fixes

aforge:
Fixes for the following:
* skipSid not set on {pres} messages initiated by the master topics as a response to client requests (e.g. user A changes permissions for user B on a topic but receives pres notifications addressed to B)
* uid may be zero when processing proxy session leave requests. This updates the `Topic.perUser` map incorrectly.

### DIFF
--- a/server/cluster.go
+++ b/server/cluster.go
@@ -1251,9 +1251,7 @@ func (sess *Session) topicProxyWriteLoop(forTopic string) {
 					// Only presence notifications may come not as a response to a client request.
 					log.Panic("cluster: message must be accompanied by the originating session information: ", srvMsg)
 				}
-				if srvMsg.skipSid != "" {
-					response.ProxyResp.SkipSid = srvMsg.skipSid
-				}
+				response.ProxyResp.SkipSid = srvMsg.skipSid
 			}
 			if copyParamsFromSession {
 				// Reply to a specific session.
--- a/server/cluster.go
+++ b/server/cluster.go
@@ -584,7 +584,7 @@ func (c *Cluster) TopicMaster(msg *ClusterReq, rejected *bool) error {
 
 	case msg.TopicMsg.MetaReq != nil:
 		if !uid.IsZero() {
-			log.Println("join setting uid = ", uid)
+			log.Println("meta setting uid = ", uid)
 			msg.CliMsg.from = uid.UserId()
 		}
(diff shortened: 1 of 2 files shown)


server/cluster.go
```diff
@@ -1245,6 +1245,15 @@ func (sess *Session) topicProxyWriteLoop(forTopic string) {
 					response.ProxyResp.OrigRequestType = ProxyRequestMeta
 				}
 				copyParamsFromSession = srvMsg.sessOverrides.sid != ""
+			} else {
+				// Copy skipSid.
+				if srvMsg.Pres == nil {
+					// Only presence notifications may come not as a response to a client request.
+					log.Panic("cluster: message must be accompanied by the originating session information: ", srvMsg)
+				}
+				if srvMsg.skipSid != "" {
```
or-else, line 1254: This check is redundant.

server/topic.go
```diff
@@ -219,11 +219,21 @@ func (t *Topic) runProxy(hub *Hub) {
 		case leave := <-t.unreg:
 			// Remove connection from topic; session may continue to function
 			log.Printf("t[%s] leave %+v", t.name, leave)
+			asUid := leave.userId
+			// Explicitly specify user id because the proxy session hosts multiple client sessions.
+			if leave.userId.IsZero() {
```
or-else, line 224: `asUid.IsZero()` for consistency?

server/topic.go
```diff
@@ -512,21 +529,25 @@ func (t *Topic) runLocal(hub *Hub) {
 						}
 					}
 					// Update user's last online timestamp & user agent
-					if err := store.Users.UpdateLastSeen(asUid, mrs.userAgent, now); err != nil {
-						log.Println(err)
+					if !proxyTerminating {
+						if err := store.Users.UpdateLastSeen(asUid, mrs.userAgent, now); err != nil {
```
or-else, line 533: Is it correct to use `asUid` vs `uid` here? I suppose they should be the same not 100% sure.

server/cluster.go
aforge: Thanks. Removed.

server/topic.go

server/topic.go
aforge: Good catch. I think you are right. We are notifying the other users of an event on behalf of `uid`, not `asUid`.

PR #434, opened 2020-04-30
Topic Proxy: parse uid from topic name for fnd topics.

aforge:


or-else:
And maybe it would be useful to give `pkt.from` a more informative name, like `pkt.asUid`. Just `from` is a bit confusing.

aforge:
Changed to `sreg.pkt.from`. Will rename `pkt.from` to `pkt.asUid` in a separate PR.

PR #435, opened 2020-05-01
Topic Proxy: create missing user entries from pres messages.

aforge:
When a `pres` message with `what=acs` contains an unknown user id, add it to the topic's `perUser`.


server/topic.go
```diff
@@ -413,23 +413,50 @@ func (t *Topic) handleCtrlBroadcast(msg *ServerComMessage) {
...
 	uid := types.ParseUserId(pres.Src)
 	dacs := pres.Acs
-	if pud, ok := t.perUser[uid]; !uid.IsZero() && ok {
+	if uid.IsZero() {
+		log.Printf("proxy topic[%s]: received acs change for invalid user id '%s'", t.name, pres.Src)
+		return
+	}
+	if pud, ok := t.perUser[uid]; ok {
+		// Existing user. Update.
 		if dacs.Want != "" {
```
or-else, line 426: This check is unnecessary because `ApplyDelta` checks it anyway.

server/topic.go
```diff
@@ -413,23 +413,50 @@ func (t *Topic) handleCtrlBroadcast(msg *ServerComMessage) {
...
 			if err := pud.modeWant.ApplyDelta(dacs.Want); err != nil {
-				log.Println("proxy topic: could not process acs change - want: ", err)
+				log.Printf("proxy topic[%s]: could not process acs change - want: %+v", t.name, err)
 			}
 		}
 		if dacs.Given != "" {
```
or-else, line 431: Also unnecessary.

server/topic.go
```diff
@@ -413,23 +413,50 @@ func (t *Topic) handleCtrlBroadcast(msg *ServerComMessage) {
...
 			}
 		}
 		t.perUser[uid] = pud
 	} else {
-		log.Println("proxy topic received acs change for unknown user ", pres.Src)
+		// New user. Create an entry in Topic's perUser.
+		want  := types.ModeNone
+		given := types.ModeNone
+
+		if dacs.Want != "" {
```
or-else, line 442: Unnecessary

server/topic.go
```diff
@@ -413,23 +413,50 @@ func (t *Topic) handleCtrlBroadcast(msg *ServerComMessage) {
...
+			if err := want.UnmarshalText([]byte(dacs.Want)); err != nil {
+				log.Printf("proxy topic [%s]: invalid acs change - want '%s'", t.name, dacs.Want)
+				return
+			}
+		}
+		if dacs.Given != "" {
```
or-else, line 448: Unnecessary check.

server/topic.go
```diff
@@ -413,23 +413,50 @@ func (t *Topic) handleCtrlBroadcast(msg *ServerComMessage) {
...
+				return
+			}
+		}
+
+		// We only check permissions on the proxy topic. The rest of bookkeeping is done on the master.
+		t.perUser[uid] = perUserData{
+			modeWant:  want,
+			modeGiven: given,
+		}
 	}
```
or-else, line 460: I would reduce code duplication:
```go
// We don't care if t.perUser[uid] exists or not. 
// If it does not exist, pud is initialized with blanks, otherwise we read pud with existing values.
// I.e. if found/if not found is not needed.
pud := t.perUser[uid]
if err := pud.modeWant.ApplyDelta(dacs.Want); err != nil {
 // report error, return
}
if err := pud.modeGiven.ApplyDelta(dacs.Given); err != nil {
 // report error, return
}
t.perUser[uid] = pud
```

aforge: Done.

server/topic.go

server/topic.go

server/topic.go

server/topic.go

server/topic.go
```diff
@@ -421,43 +421,22 @@ func (t *Topic) updateAcsFromPresMsg(pres *MsgServerPres) {
...
-			modeGiven: given,
-		}
+	// We don't care if t.perUser[uid] exists or not.
+	// If it does not exist, pud is initialized with blanks,
+	// otherwise we read pud with existing values.
+	// I.e. if found/if not found is not needed.
+	pud := t.perUser[uid]
+	if err := pud.modeWant.ApplyMutation(dacs.Want); err != nil {
+		// report error, return
+		log.Printf("proxy topic[%s]: could not process acs change - want: %+v", t.name, err)
```
or-else, line 432: You continue to execute after the error. Is it intentional?

server/topic.go
```diff
@@ -413,24 +413,30 @@ func (t *Topic) handleCtrlBroadcast(msg *ServerComMessage) {
...
+	}
+	if err := pud.modeGiven.ApplyMutation(dacs.Given); err != nil {
+		// report error, return
+		log.Printf("proxy topic[%s]: could not process acs change - given: %+v", t.name, err)
```
or-else, line 436: Same here

server/topic.go
```diff
@@ -413,24 +413,30 @@ func (t *Topic) handleCtrlBroadcast(msg *ServerComMessage) {
...
-		log.Println("proxy topic received acs change for unknown user ", pres.Src)
+	if uid.IsZero() {
+		log.Printf("proxy topic[%s]: received acs change for invalid user id '%s'", t.name, pres.Src)
+		return
+	}
+
+	// We don't care if t.perUser[uid] exists or not.
+	// If it does not exist, pud is initialized with blanks,
+	// otherwise we read pud with existing values.
+	// I.e. if found/if not found is not needed.
```
or-else, line 428: This last comment is probably not needed.

PR #441, opened 2020-05-04
Topic Proxy: terminate orphaned proxy sessions upon cluster rehashing.

aforge:


server/cluster.go
```diff
@@ -1210,6 +1210,43 @@ func (c *Cluster) invalidateProxySubs() {
...
+	}
+	globals.hub.topics.Range(func(_, v interface{}) bool {
+		topic := v.(*Topic)
+		if topic.isProxy {
+			// Topic is a proxy. Continue.
+			return true
+		}
+		for s := range topic.sessions {
+			if s.isProxy() {
+				_, originActive := activeNodeMap[s.clnode.name]
```
or-else, line 1229: More idiomatic
`if _, originActive := activeNodeMap[s.clnode.name]; !originActive { ...`

server/cluster.go
```diff
@@ -1210,6 +1210,43 @@ func (c *Cluster) invalidateProxySubs() {
...
+		return true
+	})
+	// Unsubscribe orphaned sessions from their master topics.
+	// Stop orphaned sessions.
+	for s, t := range sessions {
+		leave := &sessionLeave{
+			terminateProxyConnection: true,
+			sess:                     s,
+		}
+		t.unreg <- leave
```
or-else, line 1245: No need for variable allocation:
```
t.unreg <- &sessionLeave{
			terminateProxyConnection: true,
			sess:                     s,
		}
```
Only if you like it for readability.

server/cluster.go
aforge: Right. Changed.

server/cluster.go

PR #445, opened 2020-05-06
Proxy Topics: handling background sessions (part 1).

aforge:
Background sessions are handled by the proxy topics, not master topics.
We add such sessions to the deferred notifications list at the time
when the response to the subscription request from the master is received
and count the ticks on such sessions from that point.

When a session becomes due, we poll it off the notifications list and
send the notifications request to the master (this part is a TODO -
will come in a follow up PR).

### DIFF
--- a/server/cluster.go
+++ b/server/cluster.go
@@ -246,8 +246,8 @@ type ProxyResponse struct {
 	SkipSid string
 	// User id of the affected user.
 	Uid types.Uid
-	// Response params.
-	Params map[string]interface{}
+	// It is a response to a request from a background session.
+	IsBackground bool
 }
 
 // Handle outbound node communication: read messages from the channel, forward to remote nodes.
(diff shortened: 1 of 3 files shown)


server/cluster.go
```diff
@@ -244,6 +246,8 @@ type ProxyResponse struct {
 	SkipSid string
 	// User id of the affected user.
 	Uid types.Uid
+	// Response params.
+	Params map[string]interface{}
```
or-else, line 250: Why now use something like this?
```go
type ProxyResponseParams struct {
   background bool
} 
```

or-else: Or a pointer to it.

or-else: Or even just add `IsBackground bool` to `ProxyResponse` struct?

aforge: Sure. Changed.

PR #447, opened 2020-05-07
Topic Proxy: deferred notifications (part 2).

aforge:
Add some plumbing to make sure we use the proper user id
on the when subscribing origin sessions to proxy topics (after hearing
a word about a successful subscription from the master topic).
Also, add some data structures necessary for routing deferred notifications
requests to the master.


or-else:
Yu may want to run before committing
```bash
go fmt -w -l .
```

PR #451, opened 2020-05-11
Topic Proxy: track the list of uid proxied by proxy sessions

aforge:
This PR repurposes the no longer used Session.remoteSubs to keep track
of the proxied uids.

Also add some cleanup to make sure perUser accounting on the master topic
remains consistent in case the node hosting the proxy topic goes down.

### DIFF
--- a/server/cluster.go
+++ b/server/cluster.go
@@ -685,6 +685,8 @@ func (c *Cluster) TopicMaster(msg *ClusterReq, rejected *bool) error {
 				}
 				notifReqs = append(notifReqs, s)
 			}
+
+			sess.markRemoteSessionsForeground(notifReqs)
 			t.master <- &topicMasterRequest{deferredNotificationsRequest: notifReqs}
 		} else {
 			log.Printf("cluster: deferred notifications request for unknown topic %s", msg.RcptTo)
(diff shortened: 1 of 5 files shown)


server/cluster.go
```diff
@@ -1293,11 +1258,32 @@ func (c *Cluster) garbageCollectProxySessions(activeNodes []string) {
 	}
 }
 
+// cleanUpRemoteSubs adjusts online user counts in the master (proxied) topic
+// for a leaving proxy session.
+func (sess *Session) cleanUpRemoteSessions(topicName string) {
+	sess.remoteSessionsLock.Lock()
+	defer sess.remoteSessionsLock.Unlock()
```
or-else, line 1265: This enables a deadlock on `remoteSessionsLock` and `hub.topicGet`. Instead of `defer .. Unlock()` I would unlock before `if t := globals.hub.topicGet(topicName); t != nil {`

server/session.go
```diff
@@ -182,25 +173,24 @@ func (s *Session) unsubAll() {
...
 }
 
-func (s *Session) addRemoteSub(topic string, remoteSub *RemoteSubscription) {
-	s.remoteSubsLock.Lock()
-	defer s.remoteSubsLock.Unlock()
-
-	s.remoteSubs[topic] = remoteSub
+func (s *Session) addRemoteSession(sid string, rs *remoteSession) {
+	s.remoteSessionsLock.Lock()
+	defer s.remoteSessionsLock.Unlock()
```
or-else, line 186: `defer` is useful when you have multiple returns from the function if you need to return a value from a guarded operation. In simple cases like this it's a better practice to call `Unlock` directly because it's more efficient and uses less memory.

server/session.go
```diff
@@ -182,25 +173,24 @@ func (s *Session) unsubAll() {
...
 }
 
-func (s *Session) delRemoteSub(topic string) {
-	s.remoteSubsLock.Lock()
-	defer s.remoteSubsLock.Unlock()
-
-	delete(s.remoteSubs, topic)
+func (s *Session) delRemoteSession(sid string) {
+	s.remoteSessionsLock.Lock()
+	defer s.remoteSessionsLock.Unlock()
```
or-else, line 192: Please call `s.remoteSessionsLock.Unlock()` directly before returning from method.

server/cluster.go
aforge: True. Thanks.

server/session.go

server/session.go

PR #452, opened 2020-05-11
fix click_action in fcm messages for android

Googlom:


server/push/fcm/payload.go
```diff
@@ -246,6 +246,7 @@ func PrepareNotifications(rcpt *push.Receipt, config *AndroidConfig) []MessageDa
 							Body:        body,
 							Icon:        icon,
 							Color:       color,
+							ClickAction: config.ClickAction,
```
or-else, line 249: Please wrap it into `getClickAction` like `icon` or `color`.

or-else:
And add a comment to the config file with an example of an action.

Googlom:
> Please wrap it into `getClickAction` like `icon` or `color`.

done



> And add a comment to the config file with an example of an action.

Example already exist in config file:
https://github.com/tinode/chat/blob/8c524edf2c77d7f07e9e5088ec92442092218d2d/server/tinode.conf#L347-L348

or-else:
I can't merge - there is a conflict (`getColor` vs `getIconColor`). Please resolve it.

You may also want to update Tindroid because it won't handle click action without the following:
```xml
<intent-filter>
     <action android:name=".MessageActivity" />
     <category android:name="android.intent.category.DEFAULT" />
</intent-filter>
```
See here:
https://stackoverflow.com/questions/43372117/fcm-notification-click-action-not-working/43383425

Thanks.

Googlom:
I've fckup the branch :D
#456 'clear' PR

PR #460, opened 2020-05-17
More disambiguation and reduced amount of copying

or-else:
Made a bunch of fields in ClientComMessage serializable for intra-cluster communication and consequently they no longer have to be copied or inferred.

### DIFF
--- a/server/datamodel.go
+++ b/server/datamodel.go
@@ -597,7 +597,7 @@ type ServerComMessage struct {
 	id string
 	// Routable (expanded) name of the topic.
 	rcptto string
-	// Timestamp for consistency of timestamps in {ctrl} messages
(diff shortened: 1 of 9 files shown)


server/datamodel.go
```diff
@@ -307,17 +307,19 @@ type ClientComMessage struct {
 	Del   *MsgClientDel   `json:"del"`
 	Note  *MsgClientNote  `json:"note"`
 
+	// Internal fields, routed only within the cluster.
+
 	// Message ID denormalized
-	id string
+	Id string `json:"-"`
```
aforge, line 313: Do we need `json:"-"` here?

or-else: Yes. Otherwise it will produce json like 
`{"ctrl": {...}, "id": "123"}`

PR #462, opened 2020-05-22
Translation of README Documents into Korean

kimtaeyoung-98:
I think the README document is something that many developers read when they first encounter this project. We made README documents in Korean to attract Korean developers. If you can't reflect this contribution, please let me know why. Thank you.


or-else:
Thanks for translating Readme file. I'll merge it now.

The problem with translating the Readme is that the translation needs to be maintained. I will not be able to do it myself because I don't know the language. I'm afraid the file will become outdated quickly.

PR #472, opened 2020-06-05
Broadcast eviction to all nodes when user is deleted

or-else:
When the user is deleted, all user's sessions must be terminated on all nodes.

### DIFF
--- a/server/cluster.go
+++ b/server/cluster.go
@@ -1036,10 +1036,11 @@ func (sess *Session) clusterWriteLoop(forTopic string) {
 				// Terminating multiplexing session.
 				return
 			}
-			// A session of a remote user is being terminated.
-			// FIXME: not clear if it should be forwarded to proxy.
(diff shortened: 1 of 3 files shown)


server/cluster.go
```diff
@@ -1021,16 +1028,19 @@ func (sess *Session) clusterWriteLoop(forTopic string) {
...
-				log.Printf("cluster tp [%s]: sess.topicProxyWrite: %s", sess.sid, err.Error())
+				log.Printf("cluster master: write failed", sess.sid, err.Error())
 				return
 			}
 		case msg := <-sess.stop:
-			if msg != nil {
-				// FIXME: not clear if it should be forwarded to proxy.
-				log.Println("clusterWriteLoop stop", msg.(*ServerComMessage),
-					"to", sess.clnode.endpoint, "from", sess.sid)
+			if msg == nil {
```
aforge, line 1035: Do you need this code?

or-else: In one case the session is terminated, in the other it continues to run.

PR #475, opened 2020-06-12
Attempt to rewrite tokens in search queries.

aforge:


server/utils.go
```diff
@@ -397,6 +399,41 @@ func versionToString(vers int) string {
 	return str
 }
 
+// rewriteToken attempts to match the original token
+// against the email, telephone number or login patterns.
+// On success, prepends the token with the corresponding prefix.
+func rewriteToken(orig string) string {
+	if orig == "" || strings.HasPrefix(orig, "basic:") ||
+			strings.HasPrefix(orig, "email:") || strings.HasPrefix(orig, "tel:") {
```
or-else, line 407: I would just test for any prefix at all. I.e. if ```alphanum:``` is present then don't rewrite. If the user knows about prefixes then we should not interfere.

server/utils.go
```diff
@@ -397,6 +399,41 @@ func versionToString(vers int) string {
...
+		return "email:" + orig
+	}
+	// TODO: pass region as a param.
+	if num, err := libphonenumber.Parse(orig, "US"); err == nil {
+		// It's a phone number.
+		return "tel:" + libphonenumber.Format(num, libphonenumber.E164)
+	}
+	// Does it look like a username/login?
+	runes := []rune(orig)
+	if len(runes) >= 3 && unicode.IsLetter(runes[0]) {
```
or-else, line 421: It's fine for now but it would be much better to test against configured authenticators. Maybe add a method to authenticator to check the token and prepend with tag if it's valid. 

If you agree please add a TODO to that extent.

server/utils.go
```diff
@@ -20,11 +20,13 @@ import (
 
 	"github.com/tinode/chat/server/auth"
 	"github.com/tinode/chat/server/store/types"
+	"github.com/ttacon/libphonenumber"
 
 	"golang.org/x/crypto/acme/autocert"
 )
 
 var tagPrefixRegexp = regexp.MustCompile(`^([a-z]\w{0,5}):\S`)
+var emailRegexp = regexp.MustCompile("^[a-zA-Z0-9.!#$%&'*+/=?^_`{|}~-]+@[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?(?:\\.[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?)*$")
```
or-else, line 29: Please consider https://golang.org/pkg/net/mail/#ParseAddress

or-else:
Please don't merge it into devel. Let's release it next time. This release is already huge.

server/utils.go
```diff
@@ -397,6 +399,41 @@ func versionToString(vers int) string {
...
+func rewriteToken(orig string) string {
+	if orig == "" || strings.HasPrefix(orig, "basic:") ||
+			strings.HasPrefix(orig, "email:") || strings.HasPrefix(orig, "tel:") {
+		return orig
+	}
+	// Is it email?
+	if emailRegexp.MatchString(orig) {
+		return "email:" + orig
+	}
+	// TODO: pass region as a param.
```
or-else, line 414: I think it should be 

1. Use whatever client has provided
2. Try to infer from client's phone number or location.
3. Use value from `.conf` file.
4. Fallback to `US` as a last resort.

server/utils.go
aforge: Done.

server/utils.go
aforge: Done.

server/utils.go
aforge: Added to the TODO.

server/utils.go
aforge: Agreed. Added a todo.

aforge:
> Please don't merge it into devel. Let's release it next time. This release is already huge.

Sure. I'll keep it unmerged for now.

server/utils.go
```diff
@@ -20,13 +21,12 @@ import (
 
 	"github.com/tinode/chat/server/auth"
 	"github.com/tinode/chat/server/store/types"
-	"github.com/ttacon/libphonenumber"
+	"github.com/nyaruka/phonenumbers"
 
 	"golang.org/x/crypto/acme/autocert"
 )
 
 var tagPrefixRegexp = regexp.MustCompile(`^([a-z]\w{0,5}):\S`)
```
or-else: One more thing: the `^([a-z]\w{0,5}):` isn't quite right:
https://github.com/tinode/chat/blob/master/server/utils.go#L47

A TODO is OK for now though.

aforge: Changed to checking for any alphanumeric prefixes.

or-else:
OK to merge. Thanks!

PR #476, opened 2020-06-18
Use `AND` of `ORs` for required terms in search queries.

aforge:
When we rewrite tokens in search queries, we want to be able
to search for both the original and the rewritten terms.
Example:
- user `alice` has a tag `basic:alice`
- group `travel` is tagged `travel`

When we search for `travel` we would like to retrieve and return both results.

In order to achieve this, change the search interface as follows by
making required terms as an AND of ORs.


server/db/mongodb/adapter.go
```diff
@@ -1719,22 +1723,24 @@ func (a *adapter) getFindPipeline(req, opt []string) (map[string]struct{}, b.A)
...
+			for _, tag := range l {
+				reqTags = append(reqTags, tag)
+			}
 
-		// Filter out documents where 'tags' intersection with 'reqTags' is an empty array
-		pipeline = append(pipeline,
-			b.M{"$match": b.M{"$expr": b.M{"$ne": b.A{b.M{"$size": b.M{"$setIntersection": b.A{"$tags", reqTags}}}, 0}}}})
+			// Filter out documents where 'tags' intersection with 'reqTags' is an empty array
+			pipeline = append(pipeline,
+				b.M{"$match": b.M{"$expr": b.M{"$ne": b.A{b.M{"$size": b.M{"$setIntersection": b.A{"$tags", reqTags}}}, 0}}}})
```
or-else, line 1735: I'm not sure this is correct. The idea is to have a query like `(term OR rewritten_term) AND (term2 OR rewritten_term2)...` when it's just `term AND rewritten_term AND term2 AND rewritten_term2...`

server/db/mysql/adapter.go
```diff
@@ -1857,25 +1857,40 @@ func (a *adapter) SubsDelForUser(user t.Uid, hard bool) error {
...
+	}
+	for _, tag := range append(allReq, opt...) {
 		args = append(args, tag)
 		index[tag] = struct{}{}
 	}
 
 	query := "SELECT u.id,u.createdat,u.updatedat,u.access,u.public,u.tags,COUNT(*) AS matches " +
 		"FROM users AS u LEFT JOIN usertags AS t ON t.userid=u.id " +
-		"WHERE u.state=? AND t.tag IN (?" + strings.Repeat(",?", len(req)+len(opt)-1) + ") " +
+		"WHERE u.state=? AND t.tag IN (?" + strings.Repeat(",?", len(allReq)+len(opt)-1) + ") " +
```
or-else, line 1875: Ditto, looks like  `term AND rewritten_term AND term2 AND rewritten_term2...` instead of `(term OR rewritten_term) AND (term2 OR rewritten_term2)...`

server/db/rethinkdb/adapter.go
```diff
@@ -1534,14 +1538,16 @@ func (a *adapter) FindUsers(uid t.Uid, req, opt []string) ([]t.Subscription, err
...
-		for _, tag := range req {
-			reqTags = append(reqTags, tag)
+	if len(allReq) > 0 {
+		for _, l := range req {
+			var reqTags []interface{}
+			for _, tag := range l {
+				reqTags = append(reqTags, tag)
+			}
+			query = query.Filter(func(row rdb.Term) rdb.Term {
+				return row.Field("Tags").SetIntersection(reqTags).Count().Ne(0)
```
or-else, line 1548: Likewise. It should be `(X OR rewritten_X)` instead of `X AND rewritten_X`

server/topic.go
```diff
@@ -1791,11 +1791,16 @@ func (t *Topic) replyGetSub(sess *Session, asUid types.Uid, authLevel auth.Level
...
 			if err == nil && subs == nil && query != "" {
-				var req, opt []string
+				var req [][]string
+				var opt []string
 				if req, opt, err = parseSearchQuery(query); err == nil {
 					if len(req) > 0 || len(opt) > 0 {
 						// Check if the query contains terms that the user is not allowed to use.
-						restr, _ := stringSliceDelta(t.tags, filterRestrictedTags(append(req, opt...),
+						var allReq []string
+						for _, l := range req {
```
or-else, line 1800: This happens in a lot of places, maybe extract it into a method, like `flattedDoubleSlice` or something like that?

server/db/mongodb/adapter.go
```diff
@@ -1719,22 +1723,24 @@ func (a *adapter) getFindPipeline(req, opt []string) (map[string]struct{}, b.A)
 		b.M{"$sort": b.M{"matchedTagsCount": -1}},
 	}
 
-	if len(req) > 0 {
-		var reqTags []interface{}
-		for _, tag := range req {
-			reqTags = append(reqTags, tag)
-		}
+	if len(allReq) > 0 {
```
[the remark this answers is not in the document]
aforge: Right.

server/db/mongodb/adapter.go
aforge: It intersects rows tags with `[term, rewrittenTerm]` and checks if the intersection is empty. Which is equivalent to matching terms to `term OR rewrittenTerm`.
Am I wrong?

server/db/mysql/adapter.go
aforge: No, this line just filters all rows from the cartesian product down to those that contain any of the tags (either in `req` or in `opt`) at all.

server/db/mysql/adapter.go
```diff
@@ -1857,25 +1857,40 @@ func (a *adapter) SubsDelForUser(user t.Uid, hard bool) error {
...
+		first := true
+		for _, reqDisjunction := range req {
+			if len(reqDisjunction) > 0 {
+				if !first {
+					query += " AND"
+				} else {
+					first = false
+				}
+				// At least one of the tags must be present.
+				query += " COUNT(t.tag IN (?" + strings.Repeat(",?", len(reqDisjunction)-1) + ") OR NULL)>=1 "
```
aforge, line 1885: Note that this line changed from
`COUNT(...)>=?` to `COUNT(...)>=1`.
It is equivalent to a switch from checking a conjunction (all present) to a disjunction (at least one present.

server/db/mongodb/adapter.go
or-else: Yes, you are right, thanks. Resolved.

PR #477, opened 2020-06-21
Pass Session.countryCode to parseSearchQuery for better rewrites.

aforge:
Parse country code from the client provided language/locale.
If it fails, fall back to the default country code specified in tinode config.

### DIFF
--- a/server/session.go
+++ b/server/session.go
@@ -23,6 +23,8 @@ import (
 	"github.com/tinode/chat/server/auth"
 	"github.com/tinode/chat/server/store"
 	"github.com/tinode/chat/server/store/types"
+
+	"golang.org/x/text/language"
 )
 
 // Wait time before abandoning the outbound send operation.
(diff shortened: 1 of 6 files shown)


server/session.go
```diff
@@ -630,6 +634,15 @@ func (s *Session) hello(msg *ClientComMessage) {
 	}
 	s.deviceID = msg.Hi.DeviceID
 	s.lang = msg.Hi.Lang
+	// Try to deduce the country from the locale.
+	if tag, err := language.Parse(s.lang); err == nil {
+		if region, conf := tag.Region(); region.IsCountry() && conf >= language.High {
+			s.countryCode = region.String()
+		}
+	}
```
or-else, line 642: It may be useful to log if `s.lang` is longer than 2 chars but we are unable to extract the country. 
Longer than 2 chars because language.Parse("XX") gives confidence as 'Low'.

aforge: Done.

PR #479, opened 2020-06-26
A bit of refactoring of tag management

or-else:


server/utils.go
```diff
@@ -19,15 +19,24 @@ import (
...
 
 	"golang.org/x/crypto/acme/autocert"
 )
 
-var tagPrefixRegexp = regexp.MustCompile(`^([a-z]\w{0,5}):\S`)
-var alnumPrefixRegexp = regexp.MustCompile(`^[[:alnum:]]+:\S`)
+// Tag with prefix:
+// * prefix starts starts with an ASCII letter, contains ASCII letters, numbers, from 2 to 16 chars.
+// * tag body starts and ends with a non-space Unicode character, has 2 or more characters, otherwise not restricted.
+var tagPrefixRegexp = regexp.MustCompile(`^([a-z]\w{1,15}):(\S.*\S)$`)
```
aforge, line 32: Maybe rename prefixedTagRegexp for better readability?

server/utils.go
```diff
@@ -19,15 +19,24 @@ import (
...
+
+// Generic tag: starts and ends with a non-space Unicode character, has 2 or more characters, otherwise not restricted.
+var tagRegexp = regexp.MustCompile(`^\S.*\S$`)
+
+// Token suitable as a login: 3-16 chars, starts with a Unicode letter (class L) and contains Unicode letters (L),
+// numbers (N) and underscore.
+var basicLoginName = regexp.MustCompile(`^\pL[_\pL\pN]{2,16}$`)
```
aforge, line 39: Is it supposed to be {2,15} ?

server/utils.go
```diff
@@ -195,23 +204,38 @@ func credentialMethods(creds []MsgCredClient) []string {
 	return out
 }
 
+// isValidTag checks tag for validity: if tag contains a colon ':' then the prefix before the colon
+// must start with an ASCII letter and contain only ASCII letters, numbers and underscores.
+// The tag body may contain any combination of unicode letters, numbers, underscores and spaces.
+// The tag cannot start with a '+' or '-', cannot start or end with a space.
+func isValidTag(tag string) bool {
+	// Check if tag has a prefix.
+	if parts := tagPrefixRegexp.FindStringSubmatch(tag); len(parts) == 3 {
```
aforge, line 213: Btw, why not simply `tagPrefixRegexp.MatchString(tag)`?

server/utils.go
or-else: done

server/utils.go
or-else: done

server/utils.go
or-else: done

PR #480, opened 2020-06-28
Tag structure and length, distinction between fnd.private and fnd.public

or-else:


server/utils.go
```diff
@@ -424,42 +390,76 @@ func versionToString(vers int) string {
...
+		}
+	}
+
+	return out
+}
+
+// rewriteToken attempts to match the original token against the email, telephone number and optionally login patterns.
+// The tag is expected to be converted to lowercase.
+// On success, it prepends the token with the corresponding prefix. It returns an empty string if the tag is invalid.
+// TODO: better handling of countryCode:
```
aforge, line 421: I think this TODO is no longer relevant: we are already taking the code from the client provided locale and fallback to the config or hardcoded values.

server/utils.go
```diff
@@ -424,42 +390,76 @@ func versionToString(vers int) string {
...
-	// TODO: pass region as a param.
-	// 1. As provided by the client (e.g. as specified or inferred
-	//    from client's phone number or location).
-	// 2. Use value from .conf file.
-	// 3. Fallback to US as a last resort.
+	if addr, err := mail.ParseAddress(orig); err == nil {
+		if len([]rune(addr.Address)) < maxTagLength && addr.Address == orig {
+			return "email:" + orig
+		}
+		return ""
```
aforge, line 436: Maybe log an error here?

server/utils.go
```diff
@@ -424,42 +390,76 @@ func versionToString(vers int) string {
...
-	if basicLoginName.MatchString(orig) {
+	if withLogin && basicLoginName.MatchString(orig) {
 		return "basic:" + orig
 	}
-	return orig
+
+	if tagRegexp.MatchString(orig) {
+		return orig
+	}
+	return ""
```
aforge, line 453: Log an error here as well?
It may be useful for us to know what kind of input we skip.

docs/API.md
```diff
@@ -336,29 +336,29 @@ Message `{get what="data"}` to `me` is rejected.
...
 
-A tag is an arbitrary case-insensitive Unicode string (forced to lowercase) starting with a Unicode letter or digit. `Tags` must not contain the double quote `"`, `\u0022` but may contain spaces. `Tags` may have a prefix which serves as a namespace. The prefix is a string followed by a colon `:`, ex. prefixed phone tag `tel:+14155551212` or prefixed email tag `email:alice@example.com`. Some prefixed `tags` are optionally enforced to be unique. In that case only one user or topic may have such a tag. Certain `tags` may be forced to be immutable to the user, i.e. user's attempts to add or remove an immutable tag will be rejected by the server.
+Tag may have a prefix which serves as a namespace. The prefix is a 2-16 character string which starts with a letter [a-z] and may contain ASCII letters and numbers followed by a colon `:`, ex. prefixed phone tag `tel:+14155551212` or prefixed email tag `email:alice@example.com`. Some prefixed tags are optionally enforced to be unique. In that case only one user or topic may have such a tag. Certain tags may be forced to be immutable to the user, i.e. user's attempts to add or remove an immutable tag will be rejected by the server.
```
aforge, line 343: s/and may contain ASCII letters/and may contain lowercase ASCII letters

docs/API.md
```diff
@@ -336,29 +336,29 @@ Message `{get what="data"}` to `me` is rejected.
...
+Tinode query language is used to define search queries for finding users and topics. The query is a string containing atomic terms separated by spaces or commas. The individual query terms are matched against user's or topic's tags. The individual terms may be written in an RTL language but the query as a whole is parsed left to right. Spaces are treated as the `AND` operator, commas (as well as commas preceded and/or followed by a space) as the `OR` operator. The order of operators is ignored: all `AND` tags are grouped together, all `OR` tags are grouped together. `OR` takes precedence over `AND`: if a tag is preceded of followed by a comma, it's an `OR` tag, otherwise an `AND`. For example, `aaa bbb, ccc` (`aaa AND bbb OR ccc`) is interpreted as `(bbb OR ccc) AND aaa`.
 
-Tags containing spaces or commas must be enclosed in double quotes (`"`, `\u0022`): i.e. `"abc, def"` is treated as a single token `abc, def`. Tags must start with a Unicode letter or digit. Tags must not contain the double quote (`"`, `\u0022`).
+Query terms containing spaces must convert spaces to underscores ` ` -> `_`.
```
aforge, line 361: Maybe add an example here:
e.g. "new york" -> "new_york".

docs/API.md
```diff
@@ -369,13 +369,13 @@ Tags containing spaces or commas must be enclosed in double quotes (`"`, `\u0022
...
+Finding users by login, phone or email requires query terms to be written with prefixes, i.e. `email:alice@example.com` instead of `alice@example.com`. This may present a problem to end users because it requires them to learn the query language. Tinode solves this problem by implementing _query rewrite_ on the server: if query term (tag) does not contain a prefix, server rewrites it by adding the appropriate prefix. In queries to `fn.public` the original term is also kept (query `alice@example.com` is rewritten as `email:alice@example.com OR alice@example.com`), in queries to `fn.private` only the rewritten term is kept (`alice@example.com` is rewritten as `email:alice@example.com`). All terms that look like email, for instance, `alice@example.com` are rewritten to `email:alice@example.com OR alice@example.com`. Terms which look like phone numbers are converted to [E.164](https://en.wikipedia.org/wiki/E.164) and also rewritten as `tel:+14155551212 OR +14155551212`. In addition, in queries to `fn.public` all other unprefixed terms which look like logins are rewritten as logins: `alice` -> `basic:alice OR alice`.
```
aforge, line 378: s/fn.public/fnd.public

Also, regarding rewriting telephone numbers: mention that the country code if not specified explicitly is taken from the client-provided locale.

server/utils.go
or-else: I've added the logging, but it does not look that useful. This is a tag with spaces and such or too long tag.
I don't see an easy way to log just the unparsable phone numbers.

server/utils.go
or-else: done: changed TODO to be about inferring country code from location

server/utils.go
or-else: done

docs/API.md
or-else: done

docs/API.md
or-else: done

docs/API.md
or-else: Done

or-else:
I think we can make the phone parsing better by trying all the options: first use `hi.lang`, if fails use `config.default_country_code`, if fails use `US`.

aforge:
> I think we can make the phone parsing better by trying all the options: first use `hi.lang`, if fails use `config.default_country_code`, if fails use `US`.

That's an interesting thought. So pass a list country codes to the function?

or-else:
Yes, that's what I was thinking about.

PR #488, opened 2020-07-07
Fix minor bugs

aihturie:
Fix bug on topicsStateForUser
Fix bug cannot delete topic


aihturie:
ok sure, already done

PR #491, opened 2020-07-10
Notify old owner's access mode updates

aihturie:
Old owner's access mode updates need to be sent to subscribers when topic's owner was changed.


or-else:
Thanks

PR #494, opened 2020-07-13
Set defaultMaxMessageResults to 100 in DB adapters.

aforge:
Fetching more cannot will result in overwhelming the send queues
which effectively leads to session termination.

It's one of the problems that prevented us from running valid loadtests.

### DIFF
--- a/server/db/mongodb/adapter.go
+++ b/server/db/mongodb/adapter.go
@@ -43,8 +43,9 @@ const (
 	adpVersion  = 111
 	adapterName = "mongodb"
 
+	defaultMaxResults = 1024
 	// This is capped by the Session's send queue limit (128).
-	defaultMaxResults = 100
+	defaultMaxMessageResults = 100
 )
 
 // See https://godoc.org/go.mongodb.org/mongo-driver/mongo/options#ClientOptions for explanations.
@@ -1963,7 +1964,7 @@ func (a *adapter) MessageDeleteList(topic string, toDel *t.DelMessage) error {
 
 // MessageGetDeleted returns a list of deleted message Ids.
 func (a *adapter) MessageGetDeleted(topic string, forUser t.Uid, opts *t.QueryOpt) ([]t.DelMessage, error) {
-	var limit = a.maxMessageResults
+	var limit = a.maxResults
 	var lower, upper int
 	if opts != nil {
 		if opts.Since > 0 {
--- a/server/db/mongodb/adapter.go
+++ b/server/db/mongodb/adapter.go
@@ -24,13 +24,16 @@ import (
 
 // adapter holds MongoDB connection data.
 type adapter struct {
-	conn            *mdb.Client
-	db              *mdb.Database
-	dbName          string
-	maxResults      int
-	version         int
-	ctx             context.Context
-	useTransactions bool
+	conn              *mdb.Client
+	db                *mdb.Database
+	dbName            string
+	// Maximum number of records to return
+	maxResults        int
+	// Maximum number of message records to return
+	maxMessageResults int
+	version           int
+	ctx               context.Context
+	useTransactions   bool
 }
 
 const (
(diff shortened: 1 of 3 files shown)


aforge:
Right. Changed to use a separate maximum limit for messages.

server/db/mongodb/adapter.go
```diff
@@ -138,6 +142,10 @@ func (a *adapter) Open(jsonconfig json.RawMessage) error {
 		a.maxResults = defaultMaxResults
 	}
 
+	if a.maxMessageResults <= 0 {
+		a.maxMessageResults = defaultMaxMessageResults
```
or-else, line 146: `defaultMaxMessageResults` is not defined anywhere.

server/db/mongodb/adapter.go
```diff
@@ -40,7 +43,8 @@ const (
 	adpVersion  = 111
 	adapterName = "mongodb"
 
-	defaultMaxResults = 1024
+	// This is capped by the Session's send queue limit (128).
+	defaultMaxResults = 100
```
or-else, line 47: I think it should be left at 1024, while `defaultMaxMessageResults` should be defined as 100.

server/db/mongodb/adapter.go
```diff
@@ -1955,7 +1963,7 @@ func (a *adapter) MessageDeleteList(topic string, toDel *t.DelMessage) error {
 
 // MessageGetDeleted returns a list of deleted message Ids.
 func (a *adapter) MessageGetDeleted(topic string, forUser t.Uid, opts *t.QueryOpt) ([]t.DelMessage, error) {
-	var limit = a.maxResults
+	var limit = a.maxMessageResults
```
or-else, line 1966: It should be left as `maxResults`. Deleted ranges are collapsed into a single `meta` packet.

server/db/mysql/adapter.go
```diff
@@ -2082,7 +2090,7 @@ var dellog struct {
 
 // Get ranges of deleted messages
 func (a *adapter) MessageGetDeleted(topic string, forUser t.Uid, opts *t.QueryOpt) ([]t.DelMessage, error) {
-	var limit = a.maxResults
+	var limit = a.maxMessageResults
```
or-else, line 2093: As before, please keep `a.maxResults`.

server/db/mongodb/adapter.go
or-else: BTW, why not keep it at a slightly higher value, like 112 or 120?

server/db/rethinkdb/adapter.go
```diff
@@ -1711,7 +1720,7 @@ func (a *adapter) MessageGetAll(topic string, forUser t.Uid, opts *t.QueryOpt) (
 
 // Get ranges of deleted messages
 func (a *adapter) MessageGetDeleted(topic string, forUser t.Uid, opts *t.QueryOpt) ([]t.DelMessage, error) {
-	var limit = a.maxResults
+	var limit = a.maxMessageResults
```
or-else, line 1723: As before, no need to change.

server/db/mongodb/adapter.go
aforge: Ah, sure. Thanks for catching. Apparently, my compile command does mysql adapter only.

server/db/mongodb/adapter.go
aforge: Done.

server/db/mongodb/adapter.go
aforge: Done.

server/db/rethinkdb/adapter.go
aforge: done.

server/db/mysql/adapter.go
aforge: done.

PR #498, opened 2020-07-19
Add counters for incoming and outgoing messages.

aforge:
These are helpful in measuring performance.

### DIFF
--- a/server/sessionstore.go
+++ b/server/sessionstore.go
@@ -218,17 +218,5 @@ func NewSessionStore(lifetime time.Duration) *SessionStore {
 	statsRegisterInt("LiveSessions")
 	statsRegisterInt("TotalSessions")
 
-	statsRegisterInt("IncomingMessagesWebsockTotal")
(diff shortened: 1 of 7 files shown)


server/sessionstore.go
```diff
@@ -218,5 +218,17 @@ func NewSessionStore(lifetime time.Duration) *SessionStore {
 	statsRegisterInt("LiveSessions")
 	statsRegisterInt("TotalSessions")
 
+	statsRegisterInt("IncomingMessagesWebsockTotal")
```
or-else, line 221: It made sense to init session counters in `sessionstore.go`, but given these new initializations, maybe they should be moved to something like `statsRegisterVars()` in `stats.go`. Or maybe move to `newHub()` in `hub.go` next to 
```
statsRegisterInt("LiveTopics")
statsRegisterInt("TotalTopics")
```

aforge: Done.

PR #499, opened 2020-07-21
Channel-related functionality

or-else:
It's ~40% complete. Webapp from the https://github.com/tinode/webapp/tree/next branch can be used to test it.

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -406,11 +406,19 @@ The 'private' parameter of a P2P topic is defined by each participant individual
 
 ### Group Topics
 
-Group topics represent communication channels between multiple users. The name of a group topic is `grp` followed by a string of characters from base64 URL-encoding set. No other assumptions can be made about internal structure or length of the group name.
+Group topic represents a communication channel between multiple users. The name of a group topic is `grp` or `chn` followed by a string of characters from base64 URL-encoding set. No other assumptions can be made about internal structure or length of the group name.
 
-A group topic is created by sending a `{sub}` message with the topic field set to string `new` optionally followed by any characters, e.g. `new` or `newAbC123` are equivalent. Tinode will respond with a `{ctrl}` message with the name of the newly created topic, i.e. `{sub topic="new"}` is replied with `{ctrl topic="grpmiKBkQVXnm3P"}`. If topic creation fails, the error is reported on the original topic name, i.e. `new` or `newAbC123`. The user who created the topic becomes topic owner. Ownership can be transferred to another user with a `{set}` message but one user must remain the owner at all times.
+Group topics support limited number of subscribers (controlled by a `max_subscriber_count` parameter in configuration file) with access permissions of each subscriber managed individually. Group topics may also be enabled to support any number of read-only users - `readers`. All `readers` have the same access permissions. Group topics with enabled `readers` are called `channels`.
 
-A user joining or leaving the topic generates a `{pres}` message to all other users who are currently in the joined state with the topic.
+A group topic is created by sending a `{sub}` message with the topic field set to string `new` or `nch` optionally followed by any characters, e.g. `new` or `newAbC123` are equivalent. Tinode will respond with a `{ctrl}` message with the name of the newly created topic, i.e. `{sub topic="new"}` is replied with `{ctrl topic="grpmiKBkQVXnm3P"}`. If topic creation fails, the error is reported on the original topic name, i.e. `new` or `newAbC123`. The user who created the topic becomes topic owner. Ownership can be transferred to another user with a `{set}` message but one user must remain the owner at all times.
+
+A `channel` topic is different from the non-channel group topic in the following ways:
+
+ * Channel-enabled topic is created by sending `{sub topic="nch"}`. Sending `{sub topic="new"}` will create a group topic without enabling channel functionality.
+ * Sending `{sub topic="chnAbC123"}` will create a `reader` subscription to a channel. A non-channel topic will reject such subscription request.
+ * When searching for topics using [`fnd`](#fnd-and-tags-finding-users-and-topics), channels will show addresses with `chn` prefixes, non-channel topic will show with `grp` prefixes.
+ * Default permissions for a channel and non-channel group topics are different: channel group topic grants no permissions at all.
+ * A subscriber joining or leaving the topic generates a `{pres}` message to all other subscribers who are currently in the joined state with the topic and have appropriate permissions. Reader joining or leaving the channel generates no `{pres}` message.
 
 ### `sys` Topic
 
--- a/docs/API.md
+++ b/docs/API.md
@@ -414,11 +414,11 @@ A group topic is created by sending a `{sub}` message with the topic field set t
 
 A `channel` topic is different from the non-channel group topic in the following ways:
 
- * Channel-enabled topic is created by sending `{sub topic="nch"}`. Sending `{sub topic="new"}` will create a group topic without enabling channel functionality.
+ * Channel topic is created by sending `{sub topic="nch"}`. Sending `{sub topic="new"}` will create a group topic without enabling channel functionality.
  * Sending `{sub topic="chnAbC123"}` will create a `reader` subscription to a channel. A non-channel topic will reject such subscription request.
  * When searching for topics using [`fnd`](#fnd-and-tags-finding-users-and-topics), channels will show addresses with `chn` prefixes, non-channel topic will show with `grp` prefixes.
  * Default permissions for a channel and non-channel group topics are different: channel group topic grants no permissions at all.
- * A subscriber joining or leaving the topic generates a `{pres}` message to all other subscribers who are currently in the joined state with the topic and have appropriate permissions. Reader joining or leaving the channel generates no `{pres}` message.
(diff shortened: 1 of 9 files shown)


docs/API.md
```diff
@@ -406,11 +406,19 @@ The 'private' parameter of a P2P topic is defined by each participant individual
...
 
-A user joining or leaving the topic generates a `{pres}` message to all other users who are currently in the joined state with the topic.
+A group topic is created by sending a `{sub}` message with the topic field set to string `new` or `nch` optionally followed by any characters, e.g. `new` or `newAbC123` are equivalent. Tinode will respond with a `{ctrl}` message with the name of the newly created topic, i.e. `{sub topic="new"}` is replied with `{ctrl topic="grpmiKBkQVXnm3P"}`. If topic creation fails, the error is reported on the original topic name, i.e. `new` or `newAbC123`. The user who created the topic becomes topic owner. Ownership can be transferred to another user with a `{set}` message but one user must remain the owner at all times.
+
+A `channel` topic is different from the non-channel group topic in the following ways:
+
+ * Channel-enabled topic is created by sending `{sub topic="nch"}`. Sending `{sub topic="new"}` will create a group topic without enabling channel functionality.
```
aforge, line 417: Consider changing "Channel-enabled topic" to "Channel topic".

docs/API.md
```diff
@@ -406,11 +406,19 @@ The 'private' parameter of a P2P topic is defined by each participant individual
...
+ * Sending `{sub topic="chnAbC123"}` will create a `reader` subscription to a channel. A non-channel topic will reject such subscription request.
+ * When searching for topics using [`fnd`](#fnd-and-tags-finding-users-and-topics), channels will show addresses with `chn` prefixes, non-channel topic will show with `grp` prefixes.
+ * Default permissions for a channel and non-channel group topics are different: channel group topic grants no permissions at all.
+ * A subscriber joining or leaving the topic generates a `{pres}` message to all other subscribers who are currently in the joined state with the topic and have appropriate permissions. Reader joining or leaving the channel generates no `{pres}` message.
```
aforge, line 421: Perhaps rephase as
"A subscriber joining or leaving the topic (group or channel) generates..."

server/db/mongodb/adapter.go
```diff
@@ -1264,14 +1266,16 @@ func (a *adapter) TopicsForUser(uid t.Uid, keepDeleted bool, opts *t.QueryOpt) (
...
 				usrq = append(usrq, uid1.String())
 			}
-			topq = append(topq, sub.Topic)
+			topq = append(topq, tname)
 
 			// grp subscription
 		} else {
-			topq = append(topq, sub.Topic)
+			// Convert channel names to topic names.
+			tname = t.GrpFromChn(tname)
```
aforge, line 1274: ChnToGrp?

server/db/mongodb/adapter.go
```diff
@@ -1798,7 +1802,11 @@ func (a *adapter) FindTopics(req [][]string, opt []string) ([]t.Subscription, er
 
 		sub.CreatedAt = topic.CreatedAt
 		sub.UpdatedAt = topic.UpdatedAt
-		sub.User = topic.Id
+		if topic.UseBt {
+			sub.Topic = t.ChnFromGrp(topic.Id)
```
aforge, line 1806: GrpToChn

server/db/mysql/adapter.go
```diff
@@ -297,7 +297,7 @@ func (a *adapter) CreateDb(reset bool) error {
 			stateat   DATETIME(3),
 			touchedat DATETIME(3),
 			name      CHAR(25) NOT NULL,
-			usebt     INT DEFAULT 0,
+			usebt     TINYINT DEFAULT 0,
```
aforge, line 300: Consider leaving a comment to clarify what this bit signifies.

server/db/mysql/adapter.go
```diff
@@ -1132,9 +1132,9 @@ func (a *adapter) UserUnreadCount(uid t.Uid) (int, error) {
 // *****************************
 
 func (a *adapter) topicCreate(tx *sqlx.Tx, topic *t.Topic) error {
-	_, err := tx.Exec("INSERT INTO topics(createdat,updatedat,touchedat,state,name,owner,access,public,tags) "+
+	_, err := tx.Exec("INSERT INTO topics(createdat,updatedat,touchedat,state,name,usebt,owner,access,public,tags) "+
 		"VALUES(?,?,?,?,?,?,?,?,?)",
```
aforge, line 1136: The number of params doesn't appear to match: 10 vs 9

server/db/mysql/adapter.go
```diff
@@ -1990,14 +1993,19 @@ func (a *adapter) FindTopics(req [][]string, opt []string) ([]t.Subscription, er
...
 	for rows.Next() {
-		if err = rows.Scan(&sub.Topic, &sub.CreatedAt, &sub.UpdatedAt, &access, &public, &topicTags, &ignored); err != nil {
+		if err = rows.Scan(&sub.Topic, &sub.CreatedAt, &sub.UpdatedAt, &isChan, &access,
+			&public, &topicTags, &ignored); err != nil {
 			subs = nil
 			break
 		}
 
+		if isChan != 0 {
+			sub.Topic = t.ChnFromGrp(sub.Topic)
```
aforge, line 2007: GrpToChan?

server/session.go
```diff
@@ -451,8 +451,8 @@ func (s *Session) dispatch(msg *ClientComMessage) {
 
 // Request to subscribe to a topic
 func (s *Session) subscribe(msg *ClientComMessage) {
-	if strings.HasPrefix(msg.Original, "new") {
-		// Request to create a new named topic.
+	if strings.HasPrefix(msg.Original, "new") || strings.HasPrefix(msg.Original, "nch") {
```
aforge, line 454: Consider making this condition a utility function.
It's also present at init_topic.go:41.

server/session.go
```diff
@@ -451,8 +451,8 @@ func (s *Session) dispatch(msg *ClientComMessage) {
...
+		// Request to create a new group topic.
```
aforge, line 455: Replace "a new group topic" with "a new group/channel topic".

docs/API.md
or-else: I though about doing it that was then decided against it:
* Internally it's a group topic. It's better if docs reflect reality as closely as possible.
* There are some distinctions between a normal group topic and a channel enabled group, but the majority of features are the same.
* Client UI is different in a value of a single checkbox.

or-else: Actually, on the second thought, Channel is fine in this context as the term was explained above.

docs/API.md
or-else: I don't want to use group OR channel as it makes it appear as if channel is different from group. I added a different clarification.

server/db/mongodb/adapter.go
or-else: Done. I renamed it but missed in a couple of places.

server/db/mongodb/adapter.go
or-else: Done

server/db/mysql/adapter.go
or-else: Added comments to rethindb/schema.md. MySQL does not have comments right now. Maybe it should reference rethink instead of duplicating comments there.

server/db/mysql/adapter.go
or-else: done

server/db/mysql/adapter.go
or-else: done

server/session.go
or-else: It's only two places. I don't expect more of them. I'll add it later if needed.

server/session.go
or-else: Done

server/topic.go
```diff
@@ -1125,32 +1140,53 @@ func (t *Topic) thisUserSub(h *Hub, sess *Session, asUid types.Uid, asLvl auth.L
...
+		if isChanSub {
+			// Check if user is already subscribed.
+			var err error
+			sub, err = store.Subs.Get(pkt.Original, asUid)
+			if err != nil {
+				sess.queueOut(ErrUnknown(pkt.Id, toriginal, now))
+				return changed, err
+			}
+			tname = pkt.Original
+			userData.modeWant = types.ModeCChn
```
aforge, line 1162: Move the permission changes block to line 1141 (else clause) for better readability?

server/topic.go
```diff
@@ -1125,32 +1140,53 @@ func (t *Topic) thisUserSub(h *Hub, sess *Session, asUid types.Uid, asLvl auth.L
...
+				return changed, err
+			}
 
-		changed = true
+			changed = true
+		}
 
-		// Add user to cache.
-		usersRegisterUser(asUid, true)
+		if !isChanSub {
```
aforge, line 1186: Make this an else clause on line 1166?

server/topic.go
or-else: done

server/topic.go
or-else: User can be added to cache only when the operation has succeeded. It's not yet known in line 1166 if it has because `store.Subs.Create` may fail.

PR #503, opened 2020-07-24
Channel unsubscribe

or-else:


server/topic.go
```diff
@@ -306,24 +306,31 @@ func (t *Topic) runLocal(hub *Hub) {
 
 			// userId.IsZero() == true when the entire session is being dropped.
 			var asUid types.Uid
+			var isChan bool
```
aforge, line 309: Maybe rename `isChanReq` in order to make a clear distinction from `t.isChan`?

or-else: done `isChan` -> `asChan`

PR #504, opened 2020-07-27
Introduce histogram expvar.Var for latency tracking.

aforge:
Latencies are defined as the time between the receipt of a client request (e.g. `{pub}`) and the moment the server replies with a `{ctrl}` to this client request.
Latencies are exported as a histogram with a predefined (hardcoded) set of buckets.
Samples are recorded upon writing out `{ctrl}` messages with the specified `Id` and `Timestamp` (expected to be the timestamp of the original client request receipt) fields.

This PR currently does not address the following issues (to be implemented in follow-up PRs & commits):
1. `Timestamp` value may not be the actual original request receipt ts (e.g. https://github.com/tinode/chat/blob/11d514b58a1b5d7a3bbc67db55fa088394c73e7c/server/hub.go#L215)
2. Latencies won't be measured correctly in the cluster mode (for requests that require communicating the topic master sitting on another node).

### DIFF
--- a/server/hub.go
+++ b/server/hub.go
@@ -20,6 +20,10 @@ import (
 	"github.com/tinode/chat/server/store/types"
 )
 
+// Request latency distribution bounds (in milliseconds).
+// "var" because Go does not support array constants.
+var RequestLatencyDistribution = [...]float64 {1, 2, 3, 4, 5, 6, 8, 10, 13, 16, 20, 25, 30, 40, 50, 65, 80, 100, 130, 160, 200, 250, 300, 400, 500, 650, 800, 1000, 2000, 5000, 10000, 20000, 50000, 100000,}
+
 // Request to hub to subscribe session to topic
 type sessionJoin struct {
 	// Message, containing request details.
@@ -129,6 +133,8 @@ func newHub() *Hub {
 
 	statsRegisterInt("FileDownloadsTotal")
 	statsRegisterInt("FileUploadsTotal")
+
+	statsRegisterHistogram("RequestLatency", RequestLatencyDistribution[:]...)
 
 	go h.run()
 
--- a/server/hub.go
+++ b/server/hub.go
@@ -22,7 +22,7 @@ import (
 
 // Request latency distribution bounds (in milliseconds).
 // "var" because Go does not support array constants.
-var RequestLatencyDistribution = [...]float64 {1, 2, 3, 4, 5, 6, 8, 10, 13, 16, 20, 25, 30, 40, 50, 65, 80, 100, 130, 160, 200, 250, 300, 400, 500, 650, 800, 1000, 2000, 5000, 10000, 20000, 50000, 100000,}
+var RequestLatencyDistribution = []float64{1, 2, 3, 4, 5, 6, 8, 10, 13, 16, 20, 25, 30, 40, 50, 65, 80, 100, 130, 160, 200, 250, 300, 400, 500, 650, 800, 1000, 2000, 5000, 10000, 20000, 50000, 100000}
 
 // Request to hub to subscribe session to topic
 type sessionJoin struct {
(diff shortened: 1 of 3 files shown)


server/hub.go
```diff
@@ -20,6 +20,10 @@ import (
 	"github.com/tinode/chat/server/store/types"
 )
 
+// Request latency distribution bounds (in milliseconds).
+// "var" because Go does not support array constants.
+var RequestLatencyDistribution = [...]float64 {1, 2, 3, 4, 5, 6, 8, 10, 13, 16, 20, 25, 30, 40, 50, 65, 80, 100, 130, 160, 200, 250, 300, 400, 500, 650, 800, 1000, 2000, 5000, 10000, 20000, 50000, 100000,}
+
```
or-else, line 26: Why use an array if you later convert it to slice anyway? E.g. `[...]float64{}` -> `[]float64{}`?

or-else: Comma before `}` is not needed. `gofmt -l -w`

server/stats.go
```diff
@@ -6,18 +6,56 @@
...
+	CountPerBucket []int64   `json:"count_per_bucket"`
+	Bounds         []float64 `json:"bounds"`
+}
+
+func (h *histogram) addSample(v float64) {
+	h.Count++
+	var count *int64
+	var i int
+	var b float64
+	for i, b = range h.Bounds {
```
or-else, line 33: There are a lot of boundaries, they are sorted, and this method is called frequently. Please consider using `sort.Search`.

server/stats.go
```diff
@@ -6,18 +6,56 @@
...
+	if count == nil { // Last bucket.
+		i = len(h.Bounds)
+		count = &h.CountPerBucket[i]
+	}
+	*count++
+}
+
+func (h *histogram) String() string {
+	if r, err := json.Marshal(*h); err == nil {
+		return string(r[:])
```
or-else, line 48: Why do you need to create a copy or the byte slice here?

server/stats.go
```diff
@@ -6,18 +6,56 @@
...
+	}
+	if count == nil { // Last bucket.
+		i = len(h.Bounds)
+		count = &h.CountPerBucket[i]
+	}
+	*count++
+}
+
+func (h *histogram) String() string {
+	if r, err := json.Marshal(*h); err == nil {
```
or-else, line 47: Do you need to dereference the struct and pass it by value? It will create a copy of a potentially large struct.

server/stats.go
```diff
@@ -69,6 +116,16 @@ func statsInc(name string, val int) {
 	}
 }
 
+// Async publish a value (add a sample) to a histogram variable.
+func statsAddSample(name string, val float64) {
```
or-else, line 120: Maybe `statsAddHistSample`?

server/stats.go
```diff
@@ -49,6 +87,15 @@ func statsRegisterInt(name string) {
 	expvar.Publish(name, new(expvar.Int))
 }
 
+// Register histogram variable. `bounds` specifies histogram buckets/bins
+// (see comment next to the `histogram` struct definition).
+func statsRegisterHistogram(name string, bounds ...float64) {
```
or-else, line 92: It does not have to be variadic. You could pass the slice itself. The second parameter is always the slice.

server/hub.go
```diff
@@ -130,6 +134,8 @@ func newHub() *Hub {
 	statsRegisterInt("FileDownloadsTotal")
 	statsRegisterInt("FileUploadsTotal")
 
+	statsRegisterHistogram("RequestLatency", RequestLatencyDistribution[:]...)
```
or-else, line 137: The `statsRegisterHistogram` can just take `RequestLatencyDistribution` as a second parameter. I see no value in making `statsRegisterHistogram` variadic.

server/stats.go
aforge: Sure. It was supposed to be too generic. Of course, binary search works better here.

server/stats.go
aforge: Done.

server/stats.go
aforge: Done.

server/stats.go
aforge: Done.

PR #505, opened 2020-07-30
Latency reporting - remaining cases

aforge:
1. Properly handle incoming timestamps in the cluster mode.
2. Distinguish between server ts and incoming ts in ctrl message wrappers.

### DIFF
--- a/server/datamodel.go
+++ b/server/datamodel.go
@@ -818,12 +818,24 @@ func (src *ServerComMessage) describe() string {
 // Generators of server-side error messages {ctrl}.
 
 // NoErr indicates successful completion (200)
-func NoErr(id, topic string, serverTs, incomingReqTs time.Time) *ServerComMessage {
-	return NoErrParams(id, topic, serverTs, incomingReqTs, nil)
+func NoErr(id, topic string, ts time.Time) *ServerComMessage {
+	return NoErrParams(id, topic, ts, nil)
+}
+
+// NoErr indicates successful completion
+// with explicit server and incoming request timestamps (200)
+func NoErrExplicitTs(id, topic string, serverTs, incomingReqTs time.Time) *ServerComMessage {
+	return NoErrParamsExplicitTs(id, topic, serverTs, incomingReqTs, nil)
 }
 
 // NoErrParams indicates successful completion with additional parameters (200)
-func NoErrParams(id, topic string, serverTs, incomingReqTs time.Time, params interface{}) *ServerComMessage {
+func NoErrParams(id, topic string, ts time.Time, params interface{}) *ServerComMessage {
+	return NoErrParamsExplicitTs(id, topic, ts, ts, params)
+}
+
+// NoErrParamsExplicitTs indicates successful completion with additional parameters
+// and explicit server and incoming request timestamps (200)
+func NoErrParamsExplicitTs(id, topic string, serverTs, incomingReqTs time.Time, params interface{}) *ServerComMessage {
 	return &ServerComMessage{Ctrl: &MsgServerCtrl{
 		Id:        id,
 		Code:      http.StatusOK, // 200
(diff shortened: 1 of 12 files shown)


or-else:
This change is massive. It touches everything and we will have to maintain in perpetuity but it serves a miscellaneous use case. Is it possible to avoid changes to all error reporting functions? Maybe create two versions of every method with and without the extra parameter? Or allow passing nil when it's not needed? Shall we discuss it over VC?

aforge:
Okay, I've made changes as discussed. It still quite a big change though. Let me know what you think.

server/hub.go
```diff
@@ -230,7 +230,7 @@ func (h *Hub) run() {
 
 				log.Printf("Hub. Topic[%s] is unknown or offline", msg.RcptTo)
 
-				msg.sess.queueOut(NoErrAccepted(msg.Id, msg.RcptTo, types.TimeNow()))
+				msg.sess.queueOut(NoErrAcceptedExplicitTs(msg.Id, msg.RcptTo, types.TimeNow(), msg.Timestamp))
```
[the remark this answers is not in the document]
aforge: Yes, it's an interesting idea.
I've gone through the calls, and in many cases (especially in `topic.go`) are as follows :
`Name(id, t.original(asUid), now)`
From looking at cases, it appears `t.original(asUid)` should always be the same as `msg.Original` unless I overlooked something. What was the thought/intent behind using `t.original(asUid)` instead of `msg.Original`? Or was there nothing specific in mind?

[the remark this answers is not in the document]
or-else: > should always be the same as msg.Original

Except the case of creating a new group topic when `msg.Original` is `newXXX` or `nchXXX`.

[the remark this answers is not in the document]
or-else: > What was the thought/intent behind using t.original(asUid) instead of msg.Original?

1. Legacy
2. The case of `newXXX` or `nchXXX`.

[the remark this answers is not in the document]
or-else: The `msg.Original` was introduced very recently with the cluster refactoring.

server/datamodel.go
```diff
@@ -828,6 +828,11 @@ func NoErrExplicitTs(id, topic string, serverTs, incomingReqTs time.Time) *Serve
 	return NoErrParamsExplicitTs(id, topic, serverTs, incomingReqTs, nil)
 }
 
+// NoErrReply indicates successful completion as a reply to a client message (200).
+func NoErrReply(msg *ClientComMessage, ts time.Time) *ServerComMessage {
+	return NoErrExplicitTs(msg.Id, msg.Original, msg.Timestamp, ts)
```
or-else, line 833: I think it should be `ts, msg.Timestamp` like everywhere else. The `incomingReqTs` is `msg.Timestamp`, `serverTs` is `ts`, right?

server/datamodel.go
```diff
@@ -1021,6 +1056,12 @@ func ErrAuthRequired(id, topic string, ts time.Time) *ServerComMessage {
 		Timestamp: ts}, Id: id, Timestamp: ts}
 }
 
+// ErrAuthRequiredReply authentication required  - user must authenticate first
+// in response to a client request (401).
+func ErrAuthRequiredReply(msg *ClientComMessage) *ServerComMessage {
```
or-else, line 1061: Is there is reason not to take server time here?

server/datamodel.go
```diff
@@ -1157,13 +1229,13 @@ func ErrDuplicateCredential(id, topic string, serverTs, incomingReqTs time.Time)
 }
 
 // ErrAttachFirst must attach to topic first (409).
-func ErrAttachFirst(id, topic string, ts time.Time) *ServerComMessage {
+func ErrAttachFirst(msg *ClientComMessage) *ServerComMessage {
```
or-else, line 1232: Server ts?

server/topic.go
```diff
@@ -2186,17 +2194,18 @@ func (t *Topic) replySetTags(sess *Session, asUid types.Uid, incomingReqTs time.
...
 	if t.cat != types.TopicCatMe {
-		sess.queueOut(ErrOperationNotAllowedExplicitTs(id, t.original(asUid), now, incomingReqTs))
+		sess.queueOut(ErrOperationNotAllowedReply(msg, now))
 		return errors.New("invalid topic category for getting credentials")
 	}
 
 	screds, err := store.Users.GetAllCreds(asUid, "", false)
 	if err != nil {
-		sess.queueOut(decodeStoreErrorExplicitTs(err, id, t.original(asUid), now, incomingReqTs, nil))
+		sess.queueOut(decodeStoreErrorExplicitTs(err, id, t.original(asUid), now, msg.Timestamp, nil))
```
or-else, line 2208: `t.original(asUid)` -> `msg.Original`?

server/topic.go
```diff
@@ -2254,20 +2266,24 @@ func (t *Topic) replySetCred(sess *Session, asUid types.Uid, incomingReqTs time.
 // replyGetDel is a response to a get[what=del] request: load a list of deleted message ids, send them to
 // a session as {meta}
 // response goes to a single session rather than all sessions in a topic
-func (t *Topic) replyGetDel(sess *Session, asUid types.Uid, id string, incomingReqTs time.Time, req *MsgGetOpts) error {
+func (t *Topic) replyGetDel(sess *Session, asUid types.Uid, /*id string, incomingReqTs time.Time, req *MsgGetOpts*/msg *ClientComMessage) error {
```
or-else, line 2269: Remove comment

server/topic.go
```diff
@@ -2289,8 +2305,10 @@ func (t *Topic) replyGetDel(sess *Session, asUid types.Uid, id string, incomingR
 }
 
 // replyDelMsg deletes (soft or hard) messages in response to del.msg packet.
-func (t *Topic) replyDelMsg(sess *Session, asUid types.Uid, incomingReqTs time.Time, del *MsgClientDel) error {
+func (t *Topic) replyDelMsg(sess *Session, asUid types.Uid, /*incomingReqTs time.Time, del *MsgClientDel*/msg *ClientComMessage) error {
```
or-else, line 2308: remove comment

server/datamodel.go
aforge: Sure.

server/topic.go
aforge: Done.

server/topic.go
aforge: Done.

server/topic.go
aforge: Done.

PR #507, opened 2020-07-31
Subscribing, sending and receiving channel data messages works

or-else:


README.md
```diff
@@ -93,15 +93,13 @@ When you register a new account you are asked for an email address to send valid
 * Support for client-side data caching.
 * Ability to block unwanted communication server-side.
 * Anonymous users (important for use cases related to tech support over chat).
-* Push notifications using [FCM](https://firebase.google.com/docs/cloud-messaging/) or [TNPG](server/push/tnpg/).
```
aforge, line 96: Why remove this?

server/topic.go
```diff
@@ -223,9 +225,14 @@ func (t *Topic) userIsReader(uid types.Uid) bool {
 
 // maybeFixTopicName sets the topic field in `msg` depending on the uid.
 func (t *Topic) maybeFixTopicName(msg *ServerComMessage, uid types.Uid) {
-	if !uid.IsZero() && t.cat == types.TopicCatP2P {
+	// For zero uids we don't know the proper topic name.
+	if uid.IsZero() {
+		return
+	}
+
+	if t.cat == types.TopicCatP2P || (t.cat == types.TopicCatGrp && t.isChan) {
```
aforge, line 233: For channel topics, should we also check what the session's original subscription was?
Say, if the session subscribed to 'grpXYZ', we'll be returning 'chnXYZ' here.

server/topic.go
```diff
@@ -949,20 +956,26 @@ func (t *Topic) handleBroadcast(msg *ServerComMessage) {
...
 				}
 			}
 		}
 
 		// Topic name may be different depending on the user to which the `sess` belongs.
 		t.maybeFixTopicName(msg, pssd.uid)
+
+		// Send channel messages anonymously.
+		if pssd.isChanSub && msg.Data != nil {
+			msg.Data.From = ""
```
aforge, line 976: In this case, we don't want multiple posters on the channel?
Telegram, AFAIK, displays posters' names in channels.

README.md
or-else: I want to cull the list of features by removing expected/trivial stuff. The current list is just too long and makes it hard to see important features.

or-else: I've trimmed it down a bit more + added missing features.

server/topic.go
or-else: I would need to pass another argument to the method for a very minor efficiency improvement. It just does not seem to be worth it. The `t.original(uid)` will produce the correct topic name without knowing the original session's topic name.

server/topic.go
or-else: I checked: telegram-web always shows channel name as a poster name, telegram-android does not show the poster name at all. I did not check iOS. 
I could set `From` to the same value as the channel name but that seems redundant. Clients can do it if necessary.
I dislike the idea of showing actual poster name in channels because the name is not important. It's kind of like a newspaper.

or-else: https://telegram.org/tour/channels

> Unlike Telegram Groups, channels show the name and photo of the channel next to messages – instead of the person who posted them.

aforge: Ah, I see. I think it's the case on Telegram web client only.
E.g. https://t.me/s/my_ali_bot (see the bottom right corner of messages).

PR #512, opened 2020-08-09
Implement token authentication in loadtest.

aforge:
Tokens are saved in the log directory, and thus (a) can be re-used by other sessions and (b) are isolated between the runs.


loadtest/tinode.erl
```diff
@@ -27,3 +27,19 @@ get_rand_secret() ->
 shuffle(L) ->
   RandomList=[{rand:uniform(), X} || X <- L],
   [X || {_,X} <- lists:sort(RandomList)].
+
+%% Reads previously cached token for the specified user.
```
or-else, line 31: I would add the word 'auth' somewhere to make it clearer.

aforge: done

PR #517, opened 2020-08-14
Support for FCM topic subscriptions

or-else:


or-else:
pushgw has been changed in the other repo too. Please take a look. There is no way to set up a code review AFAIK.

server/push/fcm/push_fcm.go
```diff
@@ -101,42 +113,113 @@ func (Handler) Init(jsonconf string) error {
...
+		// Complete failure.
+		log.Println("fcm sub or upsub failed", req.Unsub, err)
+	} else {
+		// Check for partial failure.
+		handleSubErrors(resp, req.Uid, req.Devices)
+	}
+}
+
+// handlePushError processes errors returned by a call to fcm.SendAll.
+// returns true to stop further processing of other messages.
```
aforge, line 172: Perhaps, it'll be somewhat more readable to flip the return value semantics:
I read handlePushErrors() return whether errors were handled correctly and we can proceed with more pushes.

server/push/fcm/push_fcm.go
```diff
@@ -101,42 +113,113 @@ func (Handler) Init(jsonconf string) error {
...
-			} else {
-				log.Println("fcm push:", err)
-			}
+func processSubscription(req *push.ChannelReq) {
+	if len(req.Devices) == 0 {
+		return
+	}
+
+	if len(req.Devices) > subBatchSize {
+		req.Devices = req.Devices[0:subBatchSize]
```
aforge, line 152: Why not iterate over all devices and subscribe them one batch at a time?

or-else: This is expected to be called for each user individually. I can't imagine a case when a single user would have more than 1000 devices.

server/push/fcm/push_fcm.go
or-else: Done

server/push/fcm/push_fcm.go
aforge: I see. Can you add a comment explaining this please?

or-else: Done

PR #518, opened 2020-08-17
Subscribing/unsubscribing to FCM topics

or-else:
Mostly done.

Missing: 
1. Subscribe/unsubscribe to FCM topics in response to changes to P permission.
~2. Send push to user's devices when the user subscribes to a channel.~ -- done
3. Allow admins to delete channel subscriptions.
4. Maybe allow admins to add users to channels, but I'm not sure. Could be too spammy.

### DIFF
--- a/server/push/fcm/push_fcm.go
+++ b/server/push/fcm/push_fcm.go
@@ -152,6 +152,7 @@ func processSubscription(req *push.ChannelReq) {
 	if len(devices) > subBatchSize {
 		// It's extremely unlikely for a single user to have this many devices.
 		devices = devices[0:subBatchSize]
+		log.Println("fcm: user has more than", subBatchSize, "devices")
 	}
 
 	var err error
@@ -163,7 +164,7 @@ func processSubscription(req *push.ChannelReq) {
 	}
 	if err != nil {
 		// Complete failure.
-		log.Println("fcm sub or upsub failed", req.Unsub, err)
+		log.Println("fcm: sub or upsub failed", req.Unsub, err)
 	} else {
 		// Check for partial failure.
 		handleSubErrors(resp, req.Uid, devices)
--- a/server/push/fcm/push_fcm.go
+++ b/server/push/fcm/push_fcm.go
@@ -150,7 +150,6 @@ func processSubscription(req *push.ChannelReq) {
 
 	if len(req.Devices) > subBatchSize {
 		// It's extremely unlikely for a single user to have this many devices.
-		// Just clipping the list.
 		req.Devices = req.Devices[0:subBatchSize]
 	}
 
--- a/server/push/fcm/push_fcm.go
+++ b/server/push/fcm/push_fcm.go
@@ -144,28 +144,29 @@ func sendNotifications(rcpt *push.Receipt, config *configType) {
 }
 
 func processSubscription(req *push.ChannelReq) {
-	if len(req.Devices) == 0 {
+	devices := DevicesForUser(req.Uid)
+	if len(devices) == 0 {
 		return
 	}
 
-	if len(req.Devices) > subBatchSize {
+	if len(devices) > subBatchSize {
 		// It's extremely unlikely for a single user to have this many devices.
-		req.Devices = req.Devices[0:subBatchSize]
+		devices = devices[0:subBatchSize]
 	}
 
 	var err error
 	var resp *fcm.TopicManagementResponse
 	if req.Unsub {
-		resp, err = handler.client.UnsubscribeFromTopic(context.Background(), req.Devices, req.Channel)
+		resp, err = handler.client.UnsubscribeFromTopic(context.Background(), devices, req.Channel)
 	} else {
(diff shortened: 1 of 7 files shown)


server/push/tnpg/push_tnpg.go
```diff
@@ -245,14 +261,14 @@ func handlePushResponse(batch *batchResponse, messages []fcm.MessageData) {
...
-func handleSubResponse(batch *batchResponse, req *push.ChannelReq) {
+func handleSubResponse(batch *batchResponse, req *push.ChannelReq, devices []string) {
 	if batch.FailureCount <= 0 {
 		return
 	}
 
 	for _, resp := range batch.Responses {
 		// FCM documentation sucks. There is no list of possible errors so no action can be taken but logging.
-		log.Println("fcm sub/unsub error", resp.ErrorCode, req.Uid, req.Devices[resp.Index])
+		log.Println("fcm sub/unsub error", resp.ErrorCode, req.Uid, devices[resp.Index])
```
aforge, line 271: Check `resp.ErrorCode` to avoid excessive logging?

server/topic.go
```diff
@@ -1131,8 +1148,8 @@ func (t *Topic) thisUserSub(h *Hub, sess *Session, pkt *ClientComMessage, asUid
 
 	var changed bool
 
-	isChanSub := isChannel(pkt.Original)
-	if isChanSub && !t.isChan {
+	isChanSub, err := t.verifyChannelAccess(pkt.Original)
+	if err != nil {
 		// User should not be able to address non-channel topic as channel.
 		sess.queueOut(ErrNotFoundReply(pkt, now))
 		return changed, types.ErrNotFound
```
aforge, line 1155: Should we return `err` here instead of `ErrNotFound`?

server/topic.go
```diff
@@ -1242,10 +1258,17 @@ func (t *Topic) thisUserSub(h *Hub, sess *Session, pkt *ClientComMessage, asUid
 			}
 
 			changed = true
```
aforge, line 1260: It looks like it's the only place where `changed` is modified.
Otherwise, it's always false. Is it also reasonable to give it a better name, e.g. `newSub`?

server/topic.go
```diff
@@ -2392,13 +2424,20 @@ func (t *Topic) replyGetDel(sess *Session, asUid types.Uid, req *MsgGetOpts, msg
 	id := msg.Id
 	incomingReqTs := msg.Timestamp
 
+	asChan, err := t.verifyChannelAccess(toriginal)
+	if err != nil {
+		// User should not be able to address non-channel topic as channel.
+		sess.queueOut(ErrNotFoundReply(msg, now))
+		return types.ErrNotFound
```
aforge, line 2431: Return err?

server/topic.go
```diff
@@ -3211,6 +3279,18 @@ func (t *Topic) isOnline() bool {
 	return false
 }
 
+// Verifies if topic can be access by the given name.
+// Returns true if access is for channel, false if not and error if access is invalid.
+func (t *Topic) verifyChannelAccess(asTopic string) (bool, error) {
+	if isChannel(asTopic) {
```
aforge, line 3285: For better readability, format as

```
if !isChannel(asTopic) {
    return false, nil
}
if t.isChan {
    return true, nil
}
return false, types.ErrNotFound
```

server/push/tnpg/push_tnpg.go
or-else: The way I understand it, it works differently from pushes: it returns errors only for those records which produced an error. The `Index` is a reference to the original array of device IDs. In other words, `resp.ErrorCode` is always non-blank.

server/topic.go
or-else: The `err` is `types.ErrNotFound` in this case. Returning an explicit error is easier to read.

server/topic.go
or-else: The user should get `types.ErrNotFound`, it's the correct response. `err` is also `types.ErrNotFound`, but if it changes the user may get less clear error message. I would rather keep it as is.

server/topic.go
or-else: Done

server/push/fcm/push_fcm.go
```diff
@@ -144,29 +144,29 @@ func sendNotifications(rcpt *push.Receipt, config *configType) {
...
+	if len(devices) == 0 {
 		return
 	}
 
-	if len(req.Devices) > subBatchSize {
+	if len(devices) > subBatchSize {
 		// It's extremely unlikely for a single user to have this many devices.
-		// Just clipping the list.
-		req.Devices = req.Devices[0:subBatchSize]
+		devices = devices[0:subBatchSize]
```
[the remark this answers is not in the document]
or-else: done + one more place

server/topic.go
or-else: It's in two places: it's true either for created or updated subscription.

PR #520, opened 2020-08-20
Handle P access mode changes for channel readers

or-else:


server/topic.go
```diff
@@ -1258,12 +1273,21 @@ func (t *Topic) thisUserSub(h *Hub, sess *Session, pkt *ClientComMessage, asUid
 			}
 
 			changed = true
-		} else {
-			oldWant = sub.ModeWant
```
aforge, line 1262: Do you want to keep this block?
We use `oldWant` and `oldGiven` in the code below.

or-else: I moved want up to new line 1224 and given is immutable in line 1219

PR #522, opened 2020-08-25
Export response code counters.

aforge:
As CtrlCodesTotal?xx variables.

### DIFF
--- a/server/session.go
+++ b/server/session.go
@@ -254,9 +254,23 @@ func (s *Session) queueOut(msg *ServerComMessage) bool {
 	}
 
 	// Record latency only on {ctrl} messages and end-user sessions.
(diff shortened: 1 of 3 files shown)


server/session.go
```diff
@@ -254,9 +254,23 @@ func (s *Session) queueOut(msg *ServerComMessage) bool {
...
+		}
+		switch msg.Ctrl.Code / 100 {
+		case 2:
+			statsInc("CtrlCodesTotal2xx", 1)
+		case 3:
+			statsInc("CtrlCodesTotal3xx", 1)
+		case 4:
+			statsInc("CtrlCodesTotal4xx", 1)
+		case 5:
+			statsInc("CtrlCodesTotal5xx", 1)
```
or-else, line 270: Maybe `statsInc(fmt.Sprintf("CtrlCodesTotal%dxx", msg.Ctrl.Code / 100), 1)`?

or-else (approved):

server/session.go

PR #526, opened 2020-08-26
Fixed channel permissions and disabled deleting messages in channels.

or-else:
This is probably the last PR next -> channel. I'll merge channel into devel afterwards.
Things look functional with no obvious bugs. Channels can be tested with webapp from 'next'.

### DIFF
--- a/server/topic.go
+++ b/server/topic.go
@@ -2463,8 +2463,9 @@ func (t *Topic) replyGetDel(sess *Session, asUid types.Uid, req *MsgGetOpts, msg
 		return errors.New("invalid MsgGetOpts query")
 	}
 
-	// Check if the user has permission to read the topic data and the request is valid
-	if userData := t.perUser[asUid]; (userData.modeGiven & userData.modeWant).IsReader() || asChan {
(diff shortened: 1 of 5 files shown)


server/topic.go
```diff
@@ -2463,8 +2454,9 @@ func (t *Topic) replyGetDel(sess *Session, asUid types.Uid, req *MsgGetOpts, msg
 		return errors.New("invalid MsgGetOpts query")
 	}
 
-	// Check if the user has permission to read the topic data and the request is valid
-	if userData := t.perUser[asUid]; (userData.modeGiven & userData.modeWant).IsReader() || asChan {
+	// Check if the user has permission to read the topic data and the request is valid.
+	// Don't permit channel readers access deleted messages because they cannot delete messages.
```
aforge, line 2458: Channel readers cannot delete messages but should we show them gaps/placeholders for deleted messages?

or-else: This is a good point. I don't think channel readers should see the gaps, but they should be able to remove messages deleted by the publisher. I'll enable {get what=del} for channel readers. Special handling of gaps must be implemented on the client.

PR #527, opened 2020-08-26
Fail subscribe requests when the Hub's join queue is full.

aforge:
Handle subscribe responses in tsung.xml in such cases.
Also, gofmt on session.go.

### DIFF
--- a/server/session.go
+++ b/server/session.go
@@ -261,7 +261,7 @@ func (s *Session) queueOut(msg *ServerComMessage) bool {
 			statsAddHistSample("RequestLatency", float64(duration))
 		}
 		if 200 <= msg.Ctrl.Code && msg.Ctrl.Code < 600 {
-			statsInc(fmt.Sprintf("CtrlCodesTotal%dxx", msg.Ctrl.Code / 100), 1)
+			statsInc(fmt.Sprintf("CtrlCodesTotal%dxx", msg.Ctrl.Code/100), 1)
 		} else {
 			log.Println("Invalid response code: ", msg.Ctrl.Code)
 		}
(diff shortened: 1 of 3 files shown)


server/session.go
```diff
@@ -487,9 +487,15 @@ func (s *Session) subscribe(msg *ClientComMessage) {
 	if sub := s.getSub(msg.RcptTo); sub != nil {
 		s.queueOut(InfoAlreadySubscribed(msg.Id, msg.Original, msg.Timestamp))
 	} else {
-		globals.hub.join <- &sessionJoin{
+		select {
```
or-else, line 490: Let's discuss this.
If the hub is busy, then we have three options:

1. block indefinitely (original code)
2. drop immediately (this change)
3. drop after a timeout

The logic behind 1 is to let sessions wait in line until the hub becomes available. There is no harm in them waiting. It's kind of an infinite queue. The `join` is made unbuffered exactly for this reason.
 
(2) is not going to be a great user experience if `join` is unbuffered. Any kind of contention will result in dropped requests. Clients can't handle them well. We will have to implement retry logic in the clients. What is the value in dropping immediately compare to 1 or 3? At the very least the `join` must be made buffered.

(3) is a middle ground between 1 and 2. I can see some value in dropping after ~10 seconds of waiting because the context may have changed.

aforge: This change is supposed to handle the situation when the hub thread is maxed out.
If requests arrive faster than a single thread can process them (e.g. during a traffic spike) like in (1) or partly in (3), this will stall the server while the overall cpu utilization remains low.

This is precisely what I see in my loadtest runs (and note that the threading overhead is 1/4-1/3 of the overall cpu utilization). And we don't want it.

Yes, I agree we should make `join` buffered. As for timeouts (2), that's also a solution as if frees up resources eventually but we'll have to keep an extra X seconds worth of the arriving sessions (times the number of Session's threads).

or-else: OK, then let's make the `join` buffered.

aforge: Done

or-else: Please update the comment where it's declared.

or-else: The comment currently says it's unbuffered.

aforge: Done.

PR #530, opened 2020-08-27
Break Session.writeLoop() in session cleanup.

aforge:
Write loop continues to run after its host session quits. Fix it.

### DIFF
--- a/server/session.go
+++ b/server/session.go
@@ -305,6 +305,8 @@ func (s *Session) cleanUp(expired bool) {
 	s.background = false
 	s.bkgTimer.Stop()
 	s.unsubAll()
+	// Stop the write loop.
+	s.stop <- nil
 }
 
 // Message received, convert bytes to ClientComMessage and dispatch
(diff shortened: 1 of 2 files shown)


server/session.go
```diff
@@ -305,6 +305,8 @@ func (s *Session) cleanUp(expired bool) {
 	s.background = false
 	s.bkgTimer.Stop()
 	s.unsubAll()
+	// Stop the write loop.
+	s.stop <- nil
```
or-else, line 309: Could you please explain? The need for this is not obvious to me. `cleanUp` is called by the broken read loop. It also closes the underlying transport channel and that breaks the write loop. If it does not happen then it's a bug and it should be fixed.

aforge: Right. This code is for the broken read loop.
It takes the current code 9 seconds to clean up the remaining write loop (which runs in a goroutine). So with hundreds of sessions you end up with hundreds of orphaned goroutines running waiting for a timer tick in order to detect the closed transport channel and stop.
Like I mentioned in my last PR, threading overhead is showing up in the profile and we'd rather keep this overhead low. Hence, break writeLoop proactively.

or-else: OK, thanks.

or-else (approved):

Issue #529, opened 2020-08-27
too many goruntines cause very high memory usage.


hhy5861:
```
goroutine profile: total 920756
626254 @ 0x437070 0x447363 0xb3570f 0x4671c1
#	0xb3570e	main.(*Session).clusterWriteLoop+0x14e	/Users/mike/golang/poizon/duinfr/spider/server/cluster.go:1013

278952 @ 0x437070 0x447363 0xb77212 0xb65cdb 0x4671c1
#	0xb77211	main.(*Topic).runProxy+0x251	/Users/mike/golang/poizon/duinfr/spider/server/topic_proxy.go:22
#	0xb65cda	main.(*Topic).run+0x3a		/Users/mike/golang/poizon/duinfr/spider/server/topic.go:169

9478 @ 0x437070 0x447363 0xb664cd 0xb65cf8 0x4671c1
#	0xb664cc	main.(*Topic).runLocal+0x2bc	/Users/mike/golang/poizon/duinfr/spider/server/topic.go:219
#	0xb65cf7	main.(*Topic).run+0x57		/Users/mike/golang/poizon/duinfr/spider/server/topic.go:167

3289 @ 0x437070 0x447363 0xb3b952 0x4671c1
#	0xb3b951	main.(*Session).writeLoop+0x1c1	/Users/mike/golang/poizon/duinfr/spider/server/hdl_websock.go:76

2678 @ 0x437070 0x43002a 0x42f595 0x4dcfe5 0x4dde7b 0x4dde5d 0x56d7bf 0x57ff3e 0x6500d0 0x47ee31 0x65031c 0x64e8d5 0x6529fb 0x652a06 0x4fe443 0x4fe5af 0x8d2d30 0x8d528c 0x8d631f 0x8d6a9f 0xb3b5ec 0x4671c1
#	0x42f594	internal/poll.runtime_pollWait+0x54			/Users/mike/.gvm/gos/go1.14.6/src/runtime/netpoll.go:203
#	0x4dcfe4	internal/poll.(*pollDesc).wait+0x44			/Users/mike/.gvm/gos/go1.14.6/src/internal/poll/fd_poll_runtime.go:87
#	0x4dde7a	internal/poll.(*pollDesc).waitRead+0x19a		/Users/mike/.gvm/gos/go1.14.6/src/internal/poll/fd_poll_runtime.go:92
[...]

PR #533, opened 2020-08-29
Always kill proxy topic when it has no more sessions.

aforge:
It should help https://github.com/tinode/chat/issues/529


or-else (approved):
gofmt

aforge:
gofmt was noop.

PR #538, opened 2020-09-01
Set correct permissions for invited channel admins

or-else:
This is the fix for

> log in as bob
> open "Coffee" group, go to group settings
> add new member eve to the group
> log out
> log in as eve - the group doesn't show up in the list


or-else:
I have not figured out how to fix the second problem yet in a way that does not require yet another DB query.

or-else:
DB has to be reset for the fix to work.

Issue #535, opened 2020-08-30
Hold up Session cleanup until all requests completed


aforge:
Currently, a session quits whenever it receives a shutdown request.
When there are requests from this session in flight when the session is shutting down, these requests will continue to execute and may modify the server state breaking a broad range of invariants leading to memory leaks and other types of unexpected behavior.
Keep track of requests in flight, and make sure they complete before the session proceeds to the cleanup phase.

or-else:
> breaking a broad range of invariants

Can you give an example? It really should not be a problem.

aforge:
One invariant example is:
`Topic.sessions` will contain references to all *live* sessions subscribed to it.

Here's a scenario of breakage (in chronological order):
0. A join request arrives from the client.
1. Session passes it to the hub for further routing (it's now asynchronous).
2. The client terminates the session (the read/write loops are no longer running). The session cleans up and terminates all its subscriptions (but not the topic that the client intended to subscribe to in (0) ).
3. The join request gets picked up from the queue by the hub and routed to the topic.
4. The topic subscribes the session to itself (and stores a reference to the already dead session).

There's no way now for the topic in (4) to terminate because of this dangling session (the session object cannot be GC'd either).

or-else:
I see. Then maybe it would be useful to have a flag that a subscription is in flight. The session would not terminate the write loop until the answer is received. It's a bit of work but should solve the problem.

or-else:
I don't think we need to track any other message but {sub}.

or-else:
BTW, I think buffered `join` is not the cause of this problem. This scenario could have happened with unbuffered `join` just as well, except less likely. It's the early termination of the session which causes it.

PR #539, opened 2020-09-02
Session inflight sub/leave request tracking

aforge:
Should address https://github.com/tinode/chat/issues/535


server/topic_proxy.go
```diff
@@ -30,34 +30,15 @@ func (t *Topic) runProxy(hub *Hub) {
...
-				log.Println("proxy topic: route broadcast request from proxy to master failed:", err)
-			}
-			if len(t.sessions) == 0 {
-				// No more sessions attached. Start the countdown.
-				killTimer.Reset(idleProxyTopicTimeout)
+		case leave := <-t.unreg:
+			t.handleProxyLeaveRequest(leave, killTimer)
+			if leave.pkt != nil && leave.sess.inflightReqs != nil {
+				// If it's a client initiated request.
+				leave.sess.inflightReqs.Done()
```
or-else, line 41: I think it should be `Add(-1)` here. This is a multiplexing session. The leave is for a single user within the session. Also I think `handleProxyLeaveRequest` should return success/fail or the result of `t.remSession(leave.sess, asUid)`.

aforge: `Add(-1)` is equivalent to `Done()`.
Modified `t.handleProxyLeaveRequest()`.

server/session.go
```diff
@@ -520,6 +531,7 @@ func (s *Session) leave(msg *ClientComMessage) {
 		} else {
 			// Unlink from topic, topic will send a reply.
 			s.delSub(msg.RcptTo)
+			s.inflightReqs.Add(1)
```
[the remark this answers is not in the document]
or-else: Ah, you are counting both subscribes and unsubscribes. Why do you need to count unsubscribes?

server/topic.go
```diff
@@ -306,8 +306,15 @@ func (t *Topic) runLocal(hub *Hub) {
 					log.Printf("topic[%s] subscription failed %v, sid=%s", t.name, err, join.sess.sid)
 				}
 			}
+			if join.sess.inflightReqs != nil {
```
or-else, line 309: How can it be `nil`?

server/topic_proxy.go
or-else: Good point. I was under a wrong impression that `Done()` sets lock count to 0.

I don't see how `leave.sess.inflightReqs` can be zero.

or-else (approved):

server/session.go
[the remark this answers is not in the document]
aforge: Unsubscribes also modify the Topic state - so make sure we don't end up with race conditions.

server/topic.go
aforge: Multiplexed topic master (short lived) sessions don't have it yet. I should perhaps add support for these too as they modify the master topic state.

PR #540, opened 2020-09-08
Explicitly specify uid in proxy leave requests.

aforge:
Master multiplexing sessions host multiple client sessions.
Sending a leave request w/o uid will result in dropping the whole multiplexing session.
Hence, always specify the uid.

### DIFF
--- a/server/hub.go
+++ b/server/hub.go
@@ -180,7 +180,7 @@ func (h *Hub) run() {
 					xoriginal: join.pkt.Original,
 					// Indicates a proxy topic.
 					isProxy:   globals.cluster.isRemoteTopic(join.pkt.RcptTo),
-					sessions:  make(map[*Session]*perSessionData),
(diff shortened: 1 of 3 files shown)


server/hub.go
```diff
@@ -180,7 +180,7 @@ func (h *Hub) run() {
 					xoriginal: join.pkt.Original,
 					// Indicates a proxy topic.
 					isProxy:   globals.cluster.isRemoteTopic(join.pkt.RcptTo),
-					sessions:  make(map[*Session]perSessionData),
+					sessions:  make(map[*Session]*perSessionData),
```
or-else, line 183: The way I understand memory management in Go, small structs should generally not be stored as pointers.

or-else (approved):

server/hub.go
aforge: Oh, interesting. Changed back to values.

PR #542, opened 2020-09-11
Use crc32 in the Ringhash's default hashfunc implementation.

aforge:
FNV is good for uniqueness but not so good for randomnessification.
I'm switching to crc32 for now but I think we should consider something from the murmurhash family (if I'm not mistaken, G uses it for such purposes).

Some details (for reference):
The original keys are in form `tinode-{0,1,2...,N}`. We replicate and smear them across the ring as `fnv32a(%dtinode-{0,1,2,...N})` and then sort. The resulting sequence should be uniformly distributed in `[0..2^32)`.
In practice, it is not so.

If we look at the deltas between resulting hash values, we find that certain values occur too often. E.g. 16777619 which occurs 23 times which is precisely the FNV prime in [the Go implementation](https://golang.org/src/hash/fnv/fnv.go).
It is also not hard to see why it is happening: the difference between, say, `11tinode-0` and `11tinode-1` is in the last byte and if the FNV polynomial value computed up to the last iteration is small and given that xoring it with a `0` or a `1` will produce the difference in the couple of the LSBs, we end up with a relatively high chance of ending up with a difference of precisely FNV prime between the resulting hash values.


or-else (approved):

server/ringhash/ringhash.go
```diff
@@ -4,6 +4,7 @@ package ringhash
 
 import (
 	"encoding/ascii85"
+	"hash/crc32"
 	"hash/fnv"
```
or-else, line 8: Remove

aforge: FNV is still used for generating the ring signature.
I'll revisit it once I pick a good murmurhash candidate.

or-else:
This seems to be the most popular: https://github.com/spaolacci/murmur3

PR #545, opened 2020-09-12
Run topic proxy sessions' event loops on a thread pool.

aforge:
This helps to limit the resources necessary to run a separate goroutine for every multiplexing session's write event loop.

### DIFF
--- a/server/cluster.go
+++ b/server/cluster.go
@@ -916,7 +916,7 @@ func clusterInit(configString json.RawMessage, self *string) int {
 		fingerprint:     time.Now().Unix(),
 		nodes:           make(map[string]*ClusterNode),
 		// TODO: make number of workers configurable.
-		proxyEventQueue: concurrency.NewGoRoutinePool(20)}
+		proxyEventQueue: concurrency.NewGoRoutinePool(len(config.Nodes) * 5)}
 
 	var nodeNames []string
 	for _, host := range config.Nodes {
--- a/server/cluster.go
+++ b/server/cluster.go
@@ -79,6 +79,15 @@ type ClusterNode struct {
 
 	// IDs of multiplexing sessions belonging to this node.
 	msess map[string]struct{}
+
+	// Default channel for receiving responses to RPC calls issued by this node.
+	rpcDone chan *rpc.Call
+}
+
+func (n *ClusterNode) asyncRpcLoop() {
+	for call := range n.rpcDone {
+		n.handleRpcResponse(call)
+	}
 }
 
 // ClusterSess is a basic info on a remote session where the message was created.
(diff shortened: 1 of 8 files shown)


server/threadpool.go
```diff
@@ -0,0 +1,61 @@
+/******************************************************************************
+ *
+ *  Description :
+ *    A very basic and naive implementation of thread pool.
+ *
+ *****************************************************************************/
+package main
```
or-else, line 7: This is completely generic. Maybe move it to a separate package?

server/threadpool.go
```diff
@@ -0,0 +1,61 @@
...
+
+// Task represents a work task to be run on the specified thread pool.
+type Task struct {
+	// Closure to execute.
+	work func()
+}
```
or-else, line 13: The struct wrapper looks unnecessary:
```
type Task func()
```

server/cluster.go
```diff
@@ -1060,6 +1110,10 @@ func (sess *Session) clusterWriteLoop(forTopic string) {
 			// In both cases the msg does not need to be forwarded to the proxy.
 
 		case <-sess.detach:
+			log.Println("cluster: detach msg received", sess.sid)
+			return
+		default:
```
or-else, line 1115: Could you please check that this is not going to be called in a very tight loop.

server/threadpool.go
aforge: Moved to `concurrency`.

server/threadpool.go
```diff
@@ -0,0 +1,61 @@
+/******************************************************************************
+ *
+ *  Description :
+ *    A very basic and naive implementation of thread pool.
+ *
+ *****************************************************************************/
+package main
+
+// Task represents a work task to be run on the specified thread pool.
```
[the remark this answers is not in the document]
aforge: Done.

server/threadpool.go
aforge: Done.

server/cluster.go
aforge: Fair enough. Here are some time deltas between successive calls for a session:
19.1µs
1.0314ms
283.1µs
5.8601ms
15.3µs
94.7µs
175µs
1.6739ms
1.7285ms
8.9µs

Let me look more into it.

or-else: The `default` case in a `select` is run when no other case is ready. When channels are empty, the `for` will call `default` with no delay burning the CPU cycles.

or-else: It should be either a ticker or no default.

aforge: Well, yes. But there's a return statement on line 1118.
So the for loop stops once there are no more ready channels to read from and we exit the function.
What I'm more concerned about is when multiple `clusterWriteLoop` calls are scheduled in a rapid succession (hence, time deltas). In such a case, we'll have to deal with extra scheduling overhead.

or-else: Why does it take so much time?
I've got a question on architecture.Originally we had a routine per topic-node. Say, 2K topics at 20 nodes gave 40K routines at each node, right? And supposedly it was a problem? 
Right now it's 20 routines shared between all nodes. What if we pin the go routines either to the number of topics or number of nodes * X?

or-else (approved):

server/cluster.go
aforge: Sure. I'm planning to make the number of goroutines in the pool configurable.
I'll make it number of nodes * 5.

PR #546, opened 2020-09-12
Reset proxy sessions and topics for restarted node

or-else:


server/cluster.go
```diff
@@ -543,6 +558,27 @@ func (c *Cluster) UserCacheUpdate(msg *UserCacheReq, rejected *bool) error {
...
+	node := c.nodes[ping.Node]
+	if node == nil {
+		log.Println("cluster Ping from unknown node", ping.Node)
+		return nil
+	}
+
+	if node.fingerprint == 0 {
+		// This is the first connection to remote node.
+		node.fingerprint = ping.Fingerprint
+	} else if node.fingerprint == ping.Fingerprint {
```
aforge, line 572: Should it be `!=`?

or-else: Yes, fixed. Thanks.

aforge (approved):

PR #547, opened 2020-09-22
Add a basic sanity test for cluster setup.

aforge:
0. Bring up a mysql docker container.
1. Run a simple 3-node cluster.
2. Send some requests with tn-cli.


or-else (approved):
Use of ncurses reduces portability of the script. I would not use it.

aforge:
> Use of ncurses reduces portability of the script. I would not use it.

Sure. Removed ncurses constants.

PR #549, opened 2020-09-25
Run clusterWriteLoop on the master topic level.

aforge:
Remove manual scheduling.

This is a draft version of it. Please, take a look and let me know what you think.

### DIFF
--- a/server/cluster.go
+++ b/server/cluster.go
@@ -1119,11 +1119,14 @@ func (c *Cluster) gcProxySessionsForNode(node string) {
 // attached to a master topic. This function handles all the events send from
 // the master to the original sessions hosted on other nodes.
 func (t *Topic) clusterWriteLoop() {
+	cleanUp := func(sess *Session) {
+		sess.closeRPC()
+		globals.sessionStore.Delete(sess)
+		sess.unsubAll()
+	}
 	defer func() {
 		for _, sess := range t.proxiedSessions {
-			sess.closeRPC()
-			globals.sessionStore.Delete(sess)
-			sess.unsubAll()
+			cleanUp(sess)
 		}
 	}()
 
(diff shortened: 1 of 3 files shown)


server/cluster.go
```diff
@@ -1117,24 +1117,36 @@ func (c *Cluster) gcProxySessionsForNode(node string) {
 
 // clusterWriteLoop implements write loop for multiplexing (proxy) session at a node which hosts master topic.
```
or-else, line 1118: The comment looks outdated.

server/topic.go
```diff
@@ -46,6 +47,19 @@ type Topic struct {
...
+	proxiedSessions []*Session
+	// Proxied sessions' channels for the use in the topic's clusterWriteLoop:
+	// i-th session's channels (proxiedSessions[i]) are found at:
+	// proxiedChannels[i * 3 + 1] - send
+	// proxiedChannels[i * 3 + 2] - stop
+	// proxiedChannels[i * 3 + 3] - detach
+	//
+	// proxiedChannels[0] is a special-purpose channel necessary for interrupting
+	// clusterWriteLoop when sessions are added or removed.
+	proxiedChannels []reflect.SelectCase
```
or-else, line 62: I don't understand why Reflect. Why not just use a struct? It makes things hard to read without adding clear value.

or-else: Ah, I understand now. You want to use a single select to check all channels in one place. Interesting.

server/topic.go
```diff
@@ -3320,6 +3334,57 @@ func (t *Topic) subsCount() int {
 	return len(t.perUser)
 }
 
+// Adds a new multiplex proxied session to the topic's clusterWriteLoop.
+func (t *Topic) addProxiedSession(s *Session) {
```
or-else, line 3338: reflect.Select can select from up to 2^16 channels. Is it possible to have more than 2^14 sessions?

or-else: Or maybe these are multiplexing sessions, i.e. one per cluster member?

server/cluster.go

server/topic.go
aforge: Correct.

server/topic.go
aforge: The max number of multiplexing sessions we have per topic is `#number of nodes - 1`.
In practice, we should never go all the way up to `2^16`. Even 10-20 nodes is already a rather large footprint.

or-else (approved):

PR #552, opened 2020-10-05
A few fixes to allow running under high load

aforge:
1. Do not block on writing to hub's join and topic's reg channels.
2. Do not block on writing to session's send channel.
3. Add a flag to indicate that the session is terminating to prevent senders from overwhelming the session's send queue after its write loop's been broken.
4. Increase various channels' capacities.
5. Increase load in tsung.xml.


server/session.go
```diff
@@ -328,8 +353,14 @@ func (s *Session) cleanUp(expired bool) {
 
 // Message received, convert bytes to ClientComMessage and dispatch
 func (s *Session) dispatchRaw(raw []byte) {
+	now := time.Now().UTC().Round(time.Millisecond)
 	var msg ClientComMessage
 
+	if s.terminating {
+		log.Println("s.dispatch: message received on a terminating session", s.sid)
+		s.queueOut(ErrPermissionDenied("", "", now))
```
or-else, line 361: I think it should be one of 5xx errors. 4xx means the client did something wrong. 5xx means the client is fine while the server cannot handle a correct request from a client.

We have `ErrLockedReply` as 423. Consider redefining it as 503 and useing here and in case of a locked topic.

or-else (approved):

server/hub.go
```diff
@@ -209,7 +208,13 @@ func (h *Hub) run() {
 			} else {
 				// Topic found.
 				// Topic will check access rights and send appropriate {ctrl}
-				t.reg <- join
+				select {
+				case t.reg <- join:
+				default:
+					join.sess.inflightReqs.Done()
+					join.sess.queueOut(ErrUnknownReply(join.pkt, join.pkt.Timestamp))
```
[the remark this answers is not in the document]
aforge: Done.

server/session.go
```diff
@@ -328,8 +353,14 @@ func (s *Session) cleanUp(expired bool) {
 
 // Message received, convert bytes to ClientComMessage and dispatch
 func (s *Session) dispatchRaw(raw []byte) {
+	now := time.Now().UTC().Round(time.Millisecond)
```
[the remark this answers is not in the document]

server/session.go
aforge: Okay, changed ErrLocked to be a 503.

or-else:
You can merge it. I've apprved it yesterday.

server/datamodel.go
```diff
@@ -1326,22 +1326,6 @@ func ErrPolicyReply(msg *ClientComMessage, ts time.Time) *ServerComMessage {
...
-	return &ServerComMessage{Ctrl: &MsgServerCtrl{
-		Id:        id,
-		Code:      http.StatusLocked, // 423
-		Text:      "locked",
-		Topic:     topic,
-		Timestamp: serverTs}, Id: id, Timestamp: incomingReqTs}
-}
-
 // ErrLockedReply operation rejected because the topic is being deleted with explicit server and
 // incoming request timestamps in response to a client request (423).
```
or-else, line 1330: Comment should be updated 423 -> 503

server/datamodel.go
```diff
@@ -1326,22 +1326,6 @@ func ErrPolicyReply(msg *ClientComMessage, ts time.Time) *ServerComMessage {
...
 func ErrLockedReply(msg *ClientComMessage, ts time.Time) *ServerComMessage {
```
or-else, line 1331: This should be moved to line ~1400

server/datamodel.go
aforge: Done.

server/datamodel.go
aforge: Done.

PR #559, opened 2020-10-24
Make accessing Topic.{proxiedSessions,proxiedChannels} thread-safe.

aforge:
server/sanity-test.sh passes as well as a quick loadtest didn't create any trouble.

### DIFF
--- a/server/cluster.go
+++ b/server/cluster.go
@@ -30,7 +30,7 @@ const (
 // ProxyReqType is the type of proxy requests.
 type ProxyReqType int
 
-// Individual request types
+// Individual request types.
 const (
 	ProxyReqNone ProxyReqType = iota
 	ProxyReqJoin
@@ -39,6 +39,18 @@ const (
 	ProxyReqBroadcast
 	ProxyReqBgSession
 	ProxyReqMeUserAgent
+)
+
+// Proxy event types processed in the clusterWriteLoop.
+type ProxyEventType int
+
+// Individual proxy events.
+const (
+	EventSend ProxyEventType = iota
+	EventStop
+	EventDetach
+	EventContinue
+	EventAbort
 )
 
 type clusterNodeConfig struct {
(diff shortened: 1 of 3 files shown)


server/cluster.go
```diff
@@ -1098,6 +1110,34 @@ func (c *Cluster) gcProxySessionsForNode(node string) {
 	}
 }
 
+func (t *Topic) clusterSelectProxyEvent() (event ProxyEventType, s *Session, val *reflect.Value) {
```
or-else, line 1113: It looks like Go usually passes variables of the type `reflect.Value` by value rather than by reference.

server/topic.go
```diff
@@ -3342,29 +3346,44 @@ func (t *Topic) subsCount() int {
...
+	// At this point we are guaranteed to have grabbed t.proxiedLock.
 	t.proxiedSessions = append(t.proxiedSessions, s)
 	if len(t.proxiedSessions) == 1 {
 		t.proxiedChannels = make([]reflect.SelectCase, 1+3)
 		continueChan := make(chan struct{})
 		t.proxiedChannels[0] = reflect.SelectCase{Dir: reflect.SelectRecv, Chan: reflect.ValueOf(continueChan)}
-		t.proxiedChannels[1] = reflect.SelectCase{Dir: reflect.SelectRecv, Chan: reflect.ValueOf(s.send)}
-		t.proxiedChannels[2] = reflect.SelectCase{Dir: reflect.SelectRecv, Chan: reflect.ValueOf(s.stop)}
-		t.proxiedChannels[3] = reflect.SelectCase{Dir: reflect.SelectRecv, Chan: reflect.ValueOf(s.detach)}
+		t.proxiedChannels[EventSend+1] = reflect.SelectCase{Dir: reflect.SelectRecv, Chan: reflect.ValueOf(s.send)}
```
or-else, line 3366: Would if look cleaner if events are defined as `EventSend ProxyEventType = iota + 1` instead of `... = iota`?

or-else (approved):

server/topic.go
or-else: Or maybe literally = 1, = 2...

server/cluster.go
aforge: Yes, I was thinking about that as well.
I need to be able to return an empty value - hence I made it a pointer.

server/topic.go
aforge: Sure.

PR #560, opened 2020-10-25
Fix some data races.

aforge:
Something that shows up when running under tsan.

### DIFF
--- a/server/session.go
+++ b/server/session.go
@@ -257,21 +257,15 @@ func (s *Session) isCluster() bool {
 	return s.isProxy() || s.isMultiplex()
 }
 
-func (s *Session) unlockTerminating() {
-	if s == nil {
-		return
-	}
-	s.terminatingLock.Unlock()
-}
-
 // queueOut attempts to send a ServerComMessage to a session write loop; if the send buffer is full,
 // timeout is `sendTimeout`.
 func (s *Session) queueOut(msg *ServerComMessage) bool {
-	defer s.unlockTerminating()
-	if s != nil {
-		s.terminatingLock.Lock()
+	if s == nil {
+		return true
 	}
-	if s == nil || s.terminating {
+	s.terminatingLock.Lock()
+	defer s.terminatingLock.Unlock()
+	if s.terminating {
 		return true
 	}
 
@@ -363,7 +357,7 @@ func (s *Session) cleanUp(expired bool) {
 	s.unsubAll()
 	// Stop the write loop.
 	s.stopSession(nil)
-	s.unlockTerminating()
+	s.terminatingLock.Unlock()
 }
 
 // Message received, convert bytes to ClientComMessage and dispatch
--- a/server/session.go
+++ b/server/session.go
@@ -130,6 +130,8 @@ type Session struct {
 	// After this flag's been flipped to true, there must not be any more writes
 	// into the session's send channel.
 	terminating bool
+	// Guards terminating.
+	terminatingLock sync.Mutex
 
 	// Outbound mesages, buffered.
 	// The content must be serialized in format suitable for the session.
@@ -255,12 +257,24 @@ func (s *Session) isCluster() bool {
 	return s.isProxy() || s.isMultiplex()
 }
 
+func (s *Session) unlockTerminating() {
+	if s == nil {
+		return
+	}
+	s.terminatingLock.Unlock()
+}
+
 // queueOut attempts to send a ServerComMessage to a session write loop; if the send buffer is full,
 // timeout is `sendTimeout`.
 func (s *Session) queueOut(msg *ServerComMessage) bool {
+	defer s.unlockTerminating()
+	if s != nil {
+		s.terminatingLock.Lock()
+	}
 	if s == nil || s.terminating {
 		return true
 	}
+
 	if s.multi != nil {
 		// In case of a cluster we need to pass a copy of the actual session.
 		msg.sess = s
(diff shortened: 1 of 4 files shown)


server/session.go
```diff
@@ -255,12 +257,24 @@ func (s *Session) isCluster() bool {
...
+		return
+	}
+	s.terminatingLock.Unlock()
+}
+
 // queueOut attempts to send a ServerComMessage to a session write loop; if the send buffer is full,
 // timeout is `sendTimeout`.
 func (s *Session) queueOut(msg *ServerComMessage) bool {
+	defer s.unlockTerminating()
+	if s != nil {
```
or-else, line 271: `s != nil && !s.terminating` ?

server/session.go
```diff
@@ -255,12 +257,24 @@ func (s *Session) isCluster() bool {
 	return s.isProxy() || s.isMultiplex()
 }
 
+func (s *Session) unlockTerminating() {
+	if s == nil {
```
or-else, line 261: I don't see why this is needed. The `s == nil` is always checked before locking, and `nil` is never assigned to `s` while it's locked.

server/session.go
```diff
@@ -255,12 +257,24 @@ func (s *Session) isCluster() bool {
...
+	s.terminatingLock.Unlock()
+}
+
 // queueOut attempts to send a ServerComMessage to a session write loop; if the send buffer is full,
 // timeout is `sendTimeout`.
 func (s *Session) queueOut(msg *ServerComMessage) bool {
+	defer s.unlockTerminating()
+	if s != nil {
+		s.terminatingLock.Lock()
+	}
```
or-else, line 273:
```
} else {
   return
}
```

server/sessionstore.go
```diff
@@ -104,19 +104,20 @@ func (ss *SessionStore) NewSession(conn interface{}, sid string) (*Session, int)
...
 		sess.cleanUp(true)
 	}
 
-	statsSet("LiveSessions", int64(len(ss.sessCache)))
+	numSessions := len(ss.sessCache)
+	statsSet("LiveSessions", int64(numSessions))
 	statsInc("TotalSessions", 1)
 
-	return &s, len(ss.sessCache)
+	ss.lock.Unlock()
```
or-else, line 118: Why does it need to be moved here? If this is indeed needed, then there is no point in first collecting expired sessions to `expired ` and then `for _, sess := range expired`. You can call `sess.cleanUp(true)` as soon as the expired session is found. I explicitly wanted to expire session outside of the main lock to avoid locking sessions store for a potentially long time.

server/session.go
```diff
@@ -255,12 +257,24 @@ func (s *Session) isCluster() bool {
 	return s.isProxy() || s.isMultiplex()
 }
 
+func (s *Session) unlockTerminating() {
+	if s == nil {
+		return
+	}
+	s.terminatingLock.Unlock()
```
or-else, line 264: I think it's safe to repalce `s.unlockTerminating()` with `s.terminatingLock.Unlock()`.

server/session.go
```diff
@@ -255,12 +257,24 @@ func (s *Session) isCluster() bool {
...
+}
+
 // queueOut attempts to send a ServerComMessage to a session write loop; if the send buffer is full,
 // timeout is `sendTimeout`.
 func (s *Session) queueOut(msg *ServerComMessage) bool {
+	defer s.unlockTerminating()
+	if s != nil {
+		s.terminatingLock.Lock()
+	}
 	if s == nil || s.terminating {
```
or-else, line 274: Consider making this the first check in the method.

or-else (approved):

server/session.go
aforge: The race detector complains about the simultaneous access to `s.terminating`.
Given that `terminatingLock` is a member of `Session`, we need to check whether `s` is not null.

server/session.go
aforge: Changed the code to do exactly that.

server/session.go
aforge: Yup.

server/session.go
aforge: Actually, you are right. Thanks.
I didn't cleared up how `defer` works for myself. Code changed.

server/session.go

server/sessionstore.go
aforge: The race detector complains about accessing `ss.sessCache` from multiple threads (`SessionStore.NewSession` is called from hdl_websock).
So any access to `sessCache` should be mutex protected:

Write at 0x00c0002e1530 by goroutine 83:
  runtime.mapassign_faststr()
      /usr/local/go/src/runtime/map_faststr.go:202 +0x0
  main.(*SessionStore).NewSession()
      tinode/chat/server/sessionstore.go:90 +0x721
  main.serveWebSocket()
      tinode/chat/server/hdl_websock.go:169 +0x7d8

Previous read at 0x00c0002e1530 by goroutine 82:
  main.(*SessionStore).NewSession()
      tinode/chat/server/sessionstore.go:119 +0xeb4
  main.serveWebSocket()
      tinode/chat/server/hdl_websock.go:169 +0x7d8

-----------------
I'm going to move `for _, sess := range expired` outside of the lock.

or-else: Yes, that was my main concern: `for _, sess := range expired` does not need to be under the lock. `sess.cleanUp(true)` does not touch `sessionStore`.

or-else:
I've approved this PR yesterday. You can merge it.

PR #561, opened 2020-10-27
Fix a few more races.

aforge:
Oddly enough, tsan reported a few more.

### DIFF
--- a/server/cluster.go
+++ b/server/cluster.go
@@ -759,7 +759,11 @@ func (c *Cluster) isPartitioned() bool {
 		return false
 	}
 
-	return (len(c.nodes)+1)/2 >= len(c.fo.activeNodes)
+	c.fo.activeNodesLock.RLock()
+	result := (len(c.nodes)+1)/2 >= len(c.fo.activeNodes)
+	c.fo.activeNodesLock.RUnlock()
+
+	return result
 }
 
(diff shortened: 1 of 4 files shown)


or-else (approved):

or-else:
https://medium.com/a-journey-with-go/go-how-to-reduce-lock-contention-with-the-atomic-package-ba3b2664b549

aforge:
> I think some of these mutex locks can be replaced with `atomic.Store` and `atomic.Load`. The `atomic` is much faster than mutexes.
> https://medium.com/a-journey-with-go/go-how-to-reduce-lock-contention-with-the-atomic-package-ba3b2664b549

I agree. In fact, in this PR the most attractive candidate for atomics is `Session.lastAction`. Unfortunately, `time.Time` is a compound type and thus isn't compatible with atomics.

or-else:
> > I think some of these mutex locks can be replaced with `atomic.Store` and `atomic.Load`. The `atomic` is much faster than mutexes.
> > https://medium.com/a-journey-with-go/go-how-to-reduce-lock-contention-with-the-atomic-package-ba3b2664b549
> 
> I agree. In fact, in this PR the most attractive candidate for atomics is `Session.lastAction`. Unfortunately, `time.Time` is a compound type and thus isn't compatible with atomics.

We don't need `Session.lastAction` to be `Time`. It can be int64 just as easily: it's never transmitted or persisted.

PR #571, opened 2020-11-26
Fix localhost proglem in HELO/EHLO smtp command

mzch:
Fix `localhost` problem in HELO/EHLO smtp command.
Replaced `SendMail` func with a new self func.


mzch:
I refactored. I will open another PR.

PR #592, opened 2020-12-23
Split logs into info, warning, errors.

aforge:
Helps to eyeball for errors.

### DIFF
--- a/server/logs/logs.go
+++ b/server/logs/logs.go
@@ -1,0 +1,24 @@
+/******************************************************************************
+ *
+ *  Description :
+ *    Package exposes info, warning and error loggers.
+ *
+ *****************************************************************************/
+package logs
+
+import (
+	"log"
+	"os"
+)
+
+var (
+    Info    *log.Logger
+    Warning *log.Logger
+    Error   *log.Logger
+)
+
+func Init() {
+	Info = log.New(os.Stdout, "I", log.LstdFlags | log.Lshortfile)
+	Warning = log.New(os.Stdout, "W", log.LstdFlags | log.Lshortfile)
+	Error = log.New(os.Stdout, "E", log.LstdFlags | log.Lshortfile)
+}
--- a/server/logs/logs.go
+++ b/server/logs/logs.go
@@ -13,9 +13,9 @@ import (
 )
 
 var (
-    Info    *log.Logger
-    Warning *log.Logger
-    Error   *log.Logger
+	Info *log.Logger
+	Warn *log.Logger
+	Err  *log.Logger
 )
 
 func parseFlags(logFlags string) int {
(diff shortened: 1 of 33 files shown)


server/logs/logs.go
```diff
@@ -0,0 +1,24 @@
...
+	"os"
+)
+
+var (
+    Info    *log.Logger
+    Warning *log.Logger
+    Error   *log.Logger
+)
+
+func Init() {
```
or-else, line 20: The stock `log.PrintXX` writes to `os.Stderr`. Please keep it this way. Things are going to break if it's changed from `stderr` to `stdout`.

server/logs/logs.go
```diff
@@ -0,0 +1,24 @@
...
+	Info = log.New(os.Stdout, "I", log.LstdFlags | log.Lshortfile)
+	Warning = log.New(os.Stdout, "W", log.LstdFlags | log.Lshortfile)
```
or-else, line 22: I would make these less verbose, way `Warn` and `Err`. We are going to write these lines a lot.

server/logs/logs.go
aforge: Yes, thanks for pointing out. Changed (I'm playing with `stdout` locally).

server/logs/logs.go
aforge: Renamed.

server/logs/logs.go
```diff
@@ -0,0 +1,58 @@
...
+		flags = log.LstdFlags
+	}
+	return flags
+}
+
+// Initializes info, warning and error loggers given the flags
+// and the output file.
+func Init(file *os.File, logFlags string) {
+	flags := parseFlags(logFlags)
+	Info = log.New(os.Stdout, "I", flags)
```
or-else, line 55: You are not using `file` here. It's still `Stdout`.

server/logs/logs.go
```diff
@@ -0,0 +1,58 @@
...
+	}
+	if flags == 0 {
+		flags = log.LstdFlags
+	}
+	return flags
+}
+
+// Initializes info, warning and error loggers given the flags
+// and the output file.
+func Init(file *os.File, logFlags string) {
```
or-else, line 53: I would suggest to use `io.Writer` instead of `os.File`, just like `log.New` does:
https://golang.org/pkg/log/#New

`io.Writer` allows writing to sockets, `/dev/null`, and it's more flexible in general.

server/main.go
```diff
@@ -279,26 +270,36 @@ func main() {
 	var pprofUrl = flag.String("pprof_url", "", "Debugging only! URL path for exposing profiling info. Disabled if not set.")
 	flag.Parse()
 
+	logs.Init(os.Stderr, *logFlags)
```
or-else, line 273: The first parameter `os.Stderr` is not used.

server/logs/logs.go
aforge: Thanks for catching this. Fixed.

server/logs/logs.go
aforge: Changed as suggested.

server/main.go
aforge: Fixed.

or-else (approved):

PR #594, opened 2020-12-28
Detach sessions immediatelly in topic broadcast logic.

aforge:
Presently, session attached to a topic that fail to handle a broadcast
(`queueOut()` returns `false`) are detached from the topic by
sending a leave request to the topic's `unreg` channel.

This (a) incurs using a synchronization primitive and (b) may result
in dropped leave requests (if the `unreg` queue becomes full).

Change the logic to detach all "bad" sessions immediately, without
routing the leave request via the `unreg` channel.

### DIFF
--- a/server/topic.go
+++ b/server/topic.go
@@ -111,6 +111,9 @@ type Topic struct {
 
 	// Flag which tells topic lifecycle status: new, ready, paused, marked for deletion.
 	status int32
+
+	// Countdown timer for destroying the topic when there are no more attached sessions to it.
+	killTimer *time.Timer
 }
 
 // perUserData holds topic's cache of per-subscriber data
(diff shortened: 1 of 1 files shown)


server/topic.go
```diff
@@ -2707,7 +2717,7 @@ func (t *Topic) replyDelTopic(h *Hub, sess *Session, asUid types.Uid, msg *Clien
 	if t.owner != asUid {
 		// Cases 2.1.1 and 2.2
 		if t.cat != types.TopicCatP2P || t.subsCount() == 2 {
-			return t.replyLeaveUnsub(h, sess, msg, asUid)
+			return t.replyLeaveUnsub(sess, msg, asUid)
```
or-else, line 2720: Please also remove Hub from the caller `replyDelTopic`: it's no longer used.

server/topic.go
```diff
@@ -1026,20 +1036,20 @@ func (t *Topic) handleBroadcast(msg *ServerComMessage) {
...
 
 	if !t.isProxy && pushRcpt != nil {
 		// usersPush will update unread message count and send push notification.
 		usersPush(pushRcpt)
 	}
+
+	// Drop "bad" sessions.
+	for _, sess := range dropSessions {
+		// The whole session is being dropped, so sessionLeave.pkt is not set.
+		t.processLeaveRequest(&sessionLeave{sess: sess})
```
or-else, line 1051: Is this correct? `t.processLeaveRequest` calls `t.handleLeaveRequest` and that calls `t.processLeaveRequest` again.

or-else: It would also be nice to make names `processLeaveRequest` and `handleLeaveRequest` more different. And add more descriptive comments. Hard to tell what each one of them does differently from the other.

aforge: Renamed `unregisterSession`.

or-else (approved):

PR #595, opened 2021-01-03
Support for forwarding {info} to topic-offline users on "me"

or-else:
The purpose of the change is to support message previews in the chat list. The preview shows message delivery status which should be updated in real time. But the chat list is subscribed to `me` topic only. Consequently `me` should receive `{info what=recv|read}` notification for topics where the user is subscribed but not attached.

Screenshot_1609645116


server/pres.go
```diff
@@ -413,6 +413,38 @@ func (t *Topic) presSubsOffline(what string, params *presParams,
...
+		seq = read
+	} else {
+		return
+	}
+
+	user := from.UserId()
+
+	for uid, pud := range t.perUser {
+		mode := pud.modeGiven & pud.modeWant
+		if pud.deleted || !mode.IsPresencer() || !mode.IsReader() {
```
aforge, line 437: Do you want to check `pud.online` as well?

aforge (commented):
Also, can you run https://github.com/tinode/chat/blob/devel/server/sanity-test.sh please?

server/pres.go
or-else: Not here. The user may be online with some sessions but not others. I can's check session's status here. It's done at 'me'.

or-else:
`sanity-test.sh` fails to run:

```
$ ./sanity-test.sh 
+----------------------------------------------------+
|                 Tinode sanity test.                |
+----------------------------------------------------+
Timestamp = 1609779066
28081dacc1b9a20a9a15091b5dd67354bf9bd22ab243c00643fa8cf95a4ab76c
6c777dfaacd58d331f7be15e8be166a26efc7890ca0187d7522d897cc37e063f
waiting 10 seconds for mysql to come up...
[mysql] 2021/01/04 08:51:21 packets.go:36: unexpected EOF
[mysql] 2021/01/04 08:51:21 packets.go:36: unexpected EOF
[mysql] 2021/01/04 08:51:21 packets.go:36: unexpected EOF
2021/01/04 08:51:21 Database mysql 111
2021/01/04 08:51:21 Failed to init DB adapter: driver: bad connection
stop
```

I'll try to figure out why later today.

aforge (approved):

or-else:
Sanity check was failing for me for 3 reasons:

- 10 seconds wait for mysql was not always enough. I changed it to `mysqladmin ping`. 
- I did not regenerate python grpc files after changing the scheme.
- `python` is `python2` on my mac. Python2 is flaky and I don't feel a strong urge to fix it as it's outdated. I've changed python to `python3`

Also, `sanity-test.sh` was creating a docker network `tinode-net`. It was not used by anything. I removed it.

j4hangir:
> The purpose of the change is to support message previews in the chat list. The preview shows message delivery status which should be updated in real time. But the chat list is subscribed to `me` topic only. Consequently `me` should receive `{info what=recv|read}` notification for topics where the user is subscribed but not attached.
> 

Please make it into `{info what=recv|read|kp}` so when a contact is typing we get to see it right there  in the list.

Issue #440, opened 2020-05-03
Problems found in Android app


sanmiguel2019:
The project looks impressing but I can not understand some things.

Yesterday I installed Android application on 2 telephones (user1, user2) but still can not find out how both telephones can chat.

First of all I was trying to find "user2" from first phone. I searched by username, by the 6 character code that appears in section "security", by the code that starts with "usr" and looks like this - usrUig6RtjGtH. I was not able to find the second user. I was trying it on both phones. I got message saying "Topic not found".

Today in some way (I really didn't understand how) I found "user1" in phone 2. Sent a message to "user1". User1 got my message but can not read it. Over the conversation appeared message "No access to new messages" and below "Not available". I tried to tap the screen to see possible options but nothing appeared. After some time message that says that there is no access to new messages  disappeared and I was able to send a message. When I did it, the same happened for "user1" - "no access to new messages". When I return to the conversatons list, I see "You have no chats" message. I go to "find" and under "phone contacts" I see the disappeared conversation. I open the conversation, send a message. Second user gets it, however conversation's not appear in the conversations list. User have to go to contacts to see the conversation. 

Conversation periodically appears in the conversations list and then disappears again and is accessible only via contacts section. Conversations periodically are blocke with message "No access to new messages".

**Online status and push notifications don't work correctly.**

User1 access the app, then closes the app; phone screen is off. Some 3 minutes user2 still sees that user1 is online. When user2 sends a message to the user1 during this time, there is no push notification at all and user 1 can not know that he has new messages because phone screen is off.

When user closes the app or phone's screen if off, "online" status should disappear immediately.

Sometimes push notification arrive but there is no sound and I can not understand the pattern how notifications work in Tinode.

The name of received file in push notification looks uggly. It should display a thubnail of received image ir just "image", "file" as it works in all messengers, but not he file name.

Create a group. Add 1 user. Send a message. In my phone everything ok, but in the phone of added users, there s a mark "not found" below the received group message.

or-else:
What's the app version?

sanmiguel2019:
> What's the app version?

0.16.5.806 - I installed it from Play Store.

or-else:
> but still can not find out how both telephones can chat.

See here:
https://github.com/tinode/chat/blob/master/docs/API.md#query-language
Basically you need to search as `email:jdoe@example.com` or `tel:+12345678900` or `basic:user_login`. It's a known problem that will be resolved soon.

[...]

PR #596, opened 2021-01-04
Support for drafty push notifications

or-else:
The purpose of this change is to support pretty formatting of push notifications. Drafty `Preview()` strips all heavy entity data from content and shortens the text string. See https://github.com/tinode/chat/issues/440#issuecomment-624156909 "Push notifications" for an explanation.

### DIFF
--- a/server/drafty/drafty.go
+++ b/server/drafty/drafty.go
@@ -26,34 +26,54 @@ func (s *span) fromMap(in interface{}) error {
 	}
 
 	s.tp, _ = m["tp"].(string)
-	s.at = intFromNumeric(m["at"])
-	s.end = s.at + intFromNumeric(m["len"])
-	if s.end < s.at {
+	var err error
+
+	s.at, err = intFromNumeric(m["at"])
+	if err != nil {
+		return err
+	}
+
+	s.end, err = intFromNumeric(m["len"])
+	if err != nil {
+		return err
+	}
+	if s.end < 0 {
 		return errInvalidContent
 	}
+	s.end += s.at
+
 	if s.tp == "" {
-		s.key = intFromNumeric(m["key"])
+		s.key, err = intFromNumeric(m["key"])
+		if err != nil {
+			return err
+		}
+		if s.key < 0 {
+			return errInvalidContent
+		}
 	}
 
 	return nil
 }
 
-func intFromNumeric(num interface{}) int {
+func intFromNumeric(num interface{}) (int, error) {
+	if num == nil {
+		return 0, nil
+	}
 	switch i := num.(type) {
 	case int:
-		return i
+		return i, nil
 	case int16:
-		return int(i)
+		return int(i), nil
 	case int32:
-		return int(i)
+		return int(i), nil
 	case int64:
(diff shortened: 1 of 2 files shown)


server/drafty/drafty.go
```diff
@@ -19,6 +19,62 @@ type span struct {
...
+	case int32:
+		return int(i)
+	case int64:
+		return int(i)
+	case float32:
+		return int(i)
+	case float64:
+		return int(i)
+	default:
+		return 0
```
aforge, line 56: Should we fail here? E.g. return an error, or at least log an error so we know there's a problem.

Alternatively, I suggest changing the function signature to something like
`intFromNumeric(num interface{}, default int) int`
and return the `default` value in case `num` is undefined.

server/drafty/drafty.go
```diff
@@ -37,7 +93,137 @@ var tags = map[string]spanfmt{
...
+	var textLen int
+	preview := make(map[string]interface{})
+	if txtOK {
+		runes := []rune(txt)
+		textLen = len(runes)
+		if textLen > length {
+			preview["txt"] = string(runes[:length])
+			textLen = length
+		} else {
+			preview["txt"] = txt
```
aforge, line 132: Is this assignment redundant?

server/drafty/drafty.go
```diff
@@ -37,7 +93,137 @@ var tags = map[string]spanfmt{
...
+		}
+		txt = preview["txt"].(string)
```
aforge, line 134: In case the original `textLen` is less than `length` (the above else clause), this assignment is also not necessary.
That said, `preview["txt"]` needs to be modified only when the original `textLen` exceeds `length`.

server/drafty/drafty.go
```diff
@@ -37,7 +93,137 @@ var tags = map[string]spanfmt{
...
+		// Allocate space for copying styles and entities.
+		var preview_fmt []map[string]interface{}
+		var preview_ent []map[string]interface{}
+		if len(entRefs) > 0 {
+			preview_ent = make([]map[string]interface{}, len(entRefs))
+		}
+		for i := range fmt {
+			old := span{}
+			// No need to check for errors: already checked earlier.
+			old.fromMap(fmt[i])
```
aforge, line 170: You can save a `fromMap` call by storing the results from the same call on line 144 into an array.

server/drafty/drafty.go
or-else: I don't see why. It's needed.

or-else: Anyway, I simplified the whole assignment a bit.

server/drafty/drafty.go
or-else: Done

server/drafty/drafty.go
or-else: Yes, you are right. Let's fail here.

server/drafty/drafty.go
or-else: Done

aforge (approved):

PR #606, opened 2021-01-19
Channel fixes

or-else:
* `{get sub}` did not return user's own subscription in case of channel readers (no subscriptions were returned).
* `{note what=read}` did not get stored for channel readers, so the channel always had unread messages.
* There was inconsistent checking for channel vs group topic access.

### DIFF
--- a/server/topic.go
+++ b/server/topic.go
@@ -2021,7 +2021,10 @@ func (t *Topic) replyGetSub(sess *Session, asUid types.Uid, authLevel auth.Level
 		return errors.New("invalid MsgGetOpts query")
 	}
 
-	if _, err := t.verifyChannelAccess(msg.Original); err != nil {
+	var err error
+	var asChan bool
+
+	if asChan, err = t.verifyChannelAccess(msg.Original); err != nil {
 		// User should not be able to address non-channel topic as channel.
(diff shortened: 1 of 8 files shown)


or-else:
A related note: the fact that channel subscriptions are not added to `topic.perUser` seems to create some complexity. I'm thinking about maybe actually adding them bun only while there are active sessions. What do you think?

or-else:
@aforge Ping

server/topic.go
```diff
@@ -930,60 +947,83 @@ func (t *Topic) handleBroadcast(msg *ServerComMessage) {
...
+				}
 
+				upd := map[string]interface{}{}
+				if recv > 0 {
+					upd["RecvSeqId"] = recv
+				}
+				if read > 0 {
+					upd["ReadSeqId"] = read
+				}
+				if err := store.Subs.Update(topicName, asUid, upd, false); err != nil {
```
aforge, line 1001: (Perhaps that won't happen often if at all) but do we need to update subs here if `upd` is empty?

aforge (approved):

server/topic.go
or-else: It cannot be empty: `seq > 0` means wither read or recv or both are > 0.

PR #609, opened 2021-01-22
check rows.Err() in mysql adapter

gabriel-vasile:
This PR adds a check for rows.Err() in each multi-row query and removes breaks from rows.Next() loop. 

The changes follow the example from the [sql package docs](https://pkg.go.dev/database/sql#example-Rows).

### DIFF
--- a/server/db/mysql/adapter.go
+++ b/server/db/mysql/adapter.go
@@ -501,7 +501,7 @@ func (a *adapter) UpgradeDb() error {
 		// Perform database upgrade from version 107 to version 108.
 
 		// Replace default user access JRWPA with JRWPAS.
-		if _, err := a.db.Exec(`UPDATE users SET access=JSON_REPLACE(access, '$.Auth', 'JRWPAS') 
+		if _, err := a.db.Exec(`UPDATE users SET access=JSON_REPLACE(access, '$.Auth', 'JRWPAS')
 			WHERE CAST(JSON_EXTRACT(access, '$.Auth') AS CHAR) LIKE '"JRWPA"'`); err != nil {
 			return err
 		}
@@ -823,12 +823,12 @@ func (a *adapter) UserGetAll(ids ...t.Uid) ([]t.User, error) {
 	if err != nil {
 		return nil, err
 	}
+	defer rows.Close()
 
 	var user t.User
 	for rows.Next() {
-		if err = rows.StructScan(&user); err != nil {
-			users = nil
-			break
+		if err := rows.StructScan(&user); err != nil {
+			return nil, err
 		}
 
 		if user.State == t.StateDeleted {
@@ -840,7 +840,9 @@ func (a *adapter) UserGetAll(ids ...t.Uid) ([]t.User, error) {
 
 		users = append(users, user)
 	}
-	rows.Close()
+	if err := rows.Err(); err != nil {
+		return nil, err
+	}
 
 	return users, err
 }
@@ -1287,6 +1289,7 @@ func (a *adapter) TopicsForUser(uid t.Uid, keepDeleted bool, opts *t.QueryOpt) (
 	if err != nil {
 		return nil, err
 	}
+	defer rows.Close()
 
 	// Fetch subscriptions. Two queries are needed: users table (me & p2p) and topics table (p2p and grp).
 	// Prepare a list of Separate subscriptions to users vs topics
@@ -1295,8 +1298,8 @@ func (a *adapter) TopicsForUser(uid t.Uid, keepDeleted bool, opts *t.QueryOpt) (
 	topq := make([]interface{}, 0, 16)
 	usrq := make([]interface{}, 0, 16)
 	for rows.Next() {
-		if err = rows.StructScan(&sub); err != nil {
-			break
+		if err := rows.StructScan(&sub); err != nil {
+			return nil, err
 		}
 
 		tname := sub.Topic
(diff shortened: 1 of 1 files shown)


aforge (approved):
LGTM. Please, wait for @or-else to take a look before merging.
Thanks!

server/db/mysql/adapter.go
```diff
@@ -823,12 +823,12 @@ func (a *adapter) UserGetAll(ids ...t.Uid) ([]t.User, error) {
...
 		return nil, err
 	}
+	defer rows.Close()
 
 	var user t.User
 	for rows.Next() {
-		if err = rows.StructScan(&user); err != nil {
-			users = nil
-			break
+		if err := rows.StructScan(&user); err != nil {
```
or-else, line 830: The `err` is already allocated. There is no value in allocating it here again. Please use `err =` instead of `err :=`.

server/db/mysql/adapter.go
```diff
@@ -840,7 +840,9 @@ func (a *adapter) UserGetAll(ids ...t.Uid) ([]t.User, error) {
 
 		users = append(users, user)
 	}
-	rows.Close()
+	if err := rows.Err(); err != nil {
```
or-else, line 843: Likewise, there is no value in allocating `err` here when it's already allocated. Please use `err =` instead of `err :=`.

server/db/mysql/adapter.go
```diff
@@ -1295,8 +1298,8 @@ func (a *adapter) TopicsForUser(uid t.Uid, keepDeleted bool, opts *t.QueryOpt) (
 	topq := make([]interface{}, 0, 16)
 	usrq := make([]interface{}, 0, 16)
 	for rows.Next() {
-		if err = rows.StructScan(&sub); err != nil {
-			break
+		if err := rows.StructScan(&sub); err != nil {
```
or-else, line 1301: And again here.

server/db/mysql/adapter.go
```diff
@@ -1326,9 +1329,7 @@ func (a *adapter) TopicsForUser(uid t.Uid, keepDeleted bool, opts *t.QueryOpt) (
 		sub.Private = fromJSON(sub.Private)
 		join[tname] = sub
 	}
-	rows.Close()
-
-	if err != nil {
+	if err := rows.Err(); err != nil {
```
or-else, line 1332: and here

server/db/mysql/adapter.go
```diff
@@ -1518,16 +1528,19 @@ func (a *adapter) OwnTopics(uid t.Uid) ([]string, error) {
 	if err != nil {
 		return nil, err
 	}
+	defer rows.Close()
```
or-else, line 1531: In general, `defer` is less performant than a direct call. There is no need to convert all direct calls to `defer`. Please use `defer` only when it is really necessary and it indeed simplifies the code.

server/db/mysql/adapter.go
or-else: I would write it differently:
```go
for rows.Next() {
	if err = rows.StructScan(&top); err != nil {
		break
	}
.....
}
if err == nil {
	err = rows.Err()
}
row.Close()
```
It would maintain the flow with a single return and would require no `defer`.

or-else:
Regarding _The changes follow the example from the sql package docs_: examples are written by people. People have different levels of experience.

server/db/mysql/adapter.go
gabriel-vasile: There is value. It keeps the variable scope smaller.

server/db/mysql/adapter.go
gabriel-vasile: Overwriting err and breaking out of loop is not as straight forward as returning early when err is encountered.
`if err != nil { return err }` keeps the code simple and ensures you don't proceed further with spoiled results.

gabriel-vasile:
Are you suggesting those are bad examples written by inexperienced programmers? Those "people" are literally the authors of the language and the standard library.

gabriel-vasile:
Regarding your performance concerns, here are some benchmarks.
```go
func BenchmarkSingleErr(b *testing.B) {
	for n := 0; n < b.N; n++ {
		var err error
		err = errors.New("asdasd")
		err = errors.New("asdasd")
		err = errors.New("asdasd")
		err = errors.New("asdasd")
		err = errors.New("asdasd")
		_ = err
	}
}

func BenchmarkAllocErr(b *testing.B) {
	for n := 0; n < b.N; n++ {
		err1 := errors.New("asdasd")
		err2 := errors.New("asdasd")
		err3 := errors.New("asdasd")
		err4 := errors.New("asdasd")
		err5 := errors.New("asdasd")
		_, _, _, _, _ = err1, err2, err3, err4, err5
	}
}
func BenchmarkNoDefer(b *testing.B) {
	i := 0
	for n := 0; n < b.N; n++ {
		func() {
			i++
		}()
	}
}
func BenchmarkDefer(b *testing.B) {
	i := 0
	for n := 0; n < b.N; n++ {
		func() {
			i++
			defer func() {
			}()
		}()
	}
}
BenchmarkSingleErr-8   	1000000000	         0.252 ns/op
BenchmarkAllocErr-8    	1000000000	         0.246 ns/op
BenchmarkNoDefer-8     	888980235	         1.25 ns/op
BenchmarkDefer-8       	447142348	         2.88 ns/op

```

2 nanoseconds faster code without defer. Considering this communicates over the network, some milisecond delays are expected.
This makes the "optimized code" run 1/1 000 000 of a percent faster. I think code readability is more important than this invisible performance gain.

or-else:
I don't want to argue about the experience and priorities of people writing documentation examples. Let's focus on the PR

Re: benchmarks. Defer is clearly slower. The first benchmark is not obvious as compiler might have optimized the code by removing all unused assignments. Also, 

Let's just make this PR about fixing the bug instead of rewriting the code to a different style. Would you do it or should I do it?

gabriel-vasile:
I don't see myself capable of writing such optimized code, I'm afraid you'll have to do it yourself.

PR #613, opened 2021-01-27
Fix iOS push notifications.

aforge:
We must fill in aps.alert struct in order for alerts
to show up on the device.

### DIFF
--- a/server/push/fcm/payload.go
+++ b/server/push/fcm/payload.go
@@ -262,6 +262,7 @@ func PrepareNotifications(rcpt *push.Receipt, config *AndroidConfig) []MessageDa
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


server/push/fcm/payload.go
```diff
@@ -262,6 +262,8 @@ func PrepareNotifications(rcpt *push.Receipt, config *AndroidConfig) []MessageDa
 		}
 	}
 
+	titleIOS := "New message"
+	bodyIOS := data["content"]
```
or-else, line 266: It would be nice to do something smarter than hardcoding, but I suppose it's OK for now.

or-else (approved):

server/push/fcm/payload.go
aforge: Added a TODO for myself.

PR #616, opened 2021-02-22
Channels refactored

or-else:
* The `perUser` data is kept in memory as long as channel reader is attached.
* Resubscribe to channel's FCM topic when device token is obtained or refreshed (FCM subscriptions were lost before).
* Demo avatars are made larger.


server/topic.go
```diff
@@ -264,10 +273,14 @@ func (t *Topic) fixUpUserCounts(userCounts map[types.Uid]int) {
 	for uid, decrementBy := range userCounts {
```
aforge, line 273: Is this function used at all?

server/topic.go
```diff
@@ -1568,10 +1570,9 @@ func (t *Topic) thisUserSub(sess *Session, pkt *ClientComMessage, asUid types.Ui
 					"off+dis", nilPresParams, "", false)
 			}
 		}
-
-		// Apply changes.
-		t.perUser[asUid] = userData
 	}
+	// Apply changes.
+	t.perUser[asUid] = userData
```
aforge, line 1575: Aren't we supposed to avoid keeping perUserData for channel subscriptions in memory?
IIRC, the initial idea was to keep it only for owners and banned users b/c of memory issues.

server/topic.go
or-else: Removed.

server/topic.go
or-else: Yes, that was the original plan. But then I implemented the clients it did not look optimal anymore: it generated extra DB calls for trivial things like marking messages as read/received. I refactored it to keep the `perUserData` for the duration of the session. It looks like a reasonable compromise.

aforge: Okay, makes sense. So the only difference is the topic's `perUser` get loaded lazily.

server/topic.go
```diff
@@ -1350,10 +1333,12 @@ func (t *Topic) thisUserSub(sess *Session, pkt *ClientComMessage, asUid types.Ui
 				userData.modeWant = (modeWant & types.ModeCSys) | types.ModeWrite | types.ModeJoin
 			}
 		} else if asChan {
```
aforge, line 1335: Question regarding line 1400/1385 down below:
do we want `private` in `userData` for channel readers?

aforge (approved):

server/topic.go
or-else: In general we do need it: it's used to store `archived` status. Do we need to cache it for the duration of the session? Maybe not. Let me look into it.

server/topic.go
or-else: > So the only difference is the topic's perUser get loaded lazily.

Lazy load and cleared when all user's sessions disconnect.

PR #617, opened 2021-03-02
fix mysql adapter in find function

Frikker:


or-else:
Thanks. It only happens if `ONLY_FULL_GROUP_BY` is enabled.

PR #625, opened 2021-03-12
Mysql: limit the number of db connections.

aforge:
Limiting the number of connections prevents Tinode server from overwhelming
the database. It gives Tinode server a more controlled and stable behavior.
Per http://dsas.blog.klab.org/archives/2018-02/configure-sql-db.html.

### DIFF
--- a/server/db/mysql/adapter.go
+++ b/server/db/mysql/adapter.go
@@ -7,6 +7,7 @@ import (
 	"database/sql"
 	"encoding/json"
 	"errors"
+	"expvar"
 	"hash/fnv"
 	"strconv"
 	"strings"
@@ -45,8 +46,11 @@ const (
 )
 
 type configType struct {
-	DSN    string `json:"dsn,omitempty"`
-	DBName string `json:"database,omitempty"`
+	DSN                    string `json:"dsn,omitempty"`
+	DBName                 string `json:"database,omitempty"`
+	MaxOpenConns           int    `json:"max_open_conns,omitempty"`
+	MaxIdleConns           int    `json:"max_idle_conns,omitempty"`
+	ConnMaxLifetimeSeconds int    `json:"conn_max_lifetime_seconds,omitempty"`
 }
 
 // Open initializes database session
@@ -95,6 +99,20 @@ func (a *adapter) Open(jsonconfig json.RawMessage) error {
 		// Ignore missing database here. If we are initializing the database
 		// missing DB is OK.
 		err = nil
+	}
+	if err == nil {
+		if config.MaxOpenConns > 0 {
+			a.db.SetMaxOpenConns(config.MaxOpenConns)
+		}
+		if config.MaxIdleConns > 0 {
+			a.db.SetMaxIdleConns(config.MaxIdleConns)
+		}
+		if config.ConnMaxLifetimeSeconds > 0 {
+			a.db.SetConnMaxLifetime(time.Duration(config.ConnMaxLifetimeSeconds) * time.Second)
+		}
+		expvar.Publish("mysql.stats", expvar.Func(func() interface{} {
+			return a.db.Stats()
+		}))
 	}
 	return err
 }
--- a/server/db/mysql/adapter.go
+++ b/server/db/mysql/adapter.go
@@ -13,8 +13,6 @@ import (
 	"time"
 
 	ms "github.com/go-sql-driver/mysql"
-	"github.com/google/go-cmp/cmp"
-	"github.com/google/go-cmp/cmp/cmpopts"
 	"github.com/jmoiron/sqlx"
 	"github.com/tinode/chat/server/auth"
 	"github.com/tinode/chat/server/store"
@@ -47,9 +45,9 @@ const (
 )
 
 type configType struct {
-  // DB connection settings.
-  // Please, see https://pkg.go.dev/github.com/go-sql-driver/mysql#Config
-  // for the full list of fields.
+	// DB connection settings.
+	// Please, see https://pkg.go.dev/github.com/go-sql-driver/mysql#Config
+	// for the full list of fields.
 	ms.Config
 	// Deprecated.
 	DSN      string `json:"dsn,omitempty"`
@@ -78,15 +76,12 @@ func (a *adapter) Open(jsonconfig json.RawMessage) error {
 		return errors.New("mysql adapter failed to parse config: " + err.Error())
 	}
 
-	locationCmp := cmp.Comparer(func(x, y time.Location) bool {
-		return x.String() == y.String()
-	})
-	if !cmp.Equal(*defaultCfg, config.Config, locationCmp, cmpopts.IgnoreUnexported(ms.Config{})) {
+	if dsn := config.FormatDSN(); dsn != defaultCfg.FormatDSN() {
 		// MySql config is specified. Use it.
 		a.dbName = config.DBName
-		a.dsn = config.FormatDSN()
+		a.dsn = dsn
 		if config.DSN != "" || config.Database != "" {
-			return errors.New("mysql config: dsn and database fields are deprecated. Please, specify individual connection settings via mysql.Config: https://pkg.go.dev/github.com/go-sql-driver/mysql#Config")
+			return errors.New("mysql config: `dsn` and `database` fields are deprecated. Please, specify individual connection settings via mysql.Config: https://pkg.go.dev/github.com/go-sql-driver/mysql#Config")
 		}
 	} else {
 		// Otherwise, use DSN and Database to configure database connection.
--- a/server/db/mysql/adapter.go
+++ b/server/db/mysql/adapter.go
@@ -7,13 +7,14 @@ import (
 	"database/sql"
 	"encoding/json"
 	"errors"
-	"fmt"
 	"hash/fnv"
 	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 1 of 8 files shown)


server/db/mysql/adapter.go
```diff
@@ -45,8 +46,11 @@ const (
 )
 
 type configType struct {
-	DSN    string `json:"dsn,omitempty"`
-	DBName string `json:"database,omitempty"`
+	DSN                    string `json:"dsn,omitempty"`
```
or-else, line 49: Some connection parameters are specified in the DSN, some as explicit variables. Maybe get rid of the DSN altogether and use just the explicit variables? For example, construct DSN using https://pkg.go.dev/github.com/go-sql-driver/mysql#Config.FormatDSN

I don't insist. It just seems more consistent.

aforge: These new params appear to be mysql driver specific settings.
The MySql specific DSN config does not contain them: https://pkg.go.dev/github.com/go-sql-driver/mysql#Config

server/db/mysql/adapter.go
```diff
@@ -96,6 +100,20 @@ func (a *adapter) Open(jsonconfig json.RawMessage) error {
...
+		if config.MaxOpenConns > 0 {
+			a.db.SetMaxOpenConns(config.MaxOpenConns)
+		}
+		if config.MaxIdleConns > 0 {
+			a.db.SetMaxIdleConns(config.MaxIdleConns)
+		}
+		if config.ConnMaxLifetimeSeconds > 0 {
+			a.db.SetConnMaxLifetime(time.Duration(config.ConnMaxLifetimeSeconds) * time.Second)
+		}
+		expvar.Publish("mysql.stats", expvar.Func(func() interface{} {
```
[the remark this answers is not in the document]
aforge: Ack.

server/db/mysql/adapter.go
```diff
@@ -7,6 +7,7 @@ import (
 	"database/sql"
 	"encoding/json"
 	"errors"
+	"expvar"
```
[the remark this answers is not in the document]
aforge: Okay. Makes sense.

How about I add a new API function to `Store` which will return a stats object
and export it in expvar in `stats.go`?

server/db/mysql/adapter.go
aforge: One thing we can do though is define a field of type `Config` and format it into DSN.
What do you think?

```go
type configType struct {
	DSNConfig Config `dsn_config`
        DBName string `json:"database,omitempty"`
        // Connection pool params.
	MaxOpenConns           int    `json:"max_open_conns,omitempty"`
	MaxIdleConns           int    `json:"max_idle_conns,omitempty"`
	ConnMaxLifetimeSeconds int    `json:"conn_max_lifetime_seconds,omitempty"`
}
```

server/db/mysql/adapter.go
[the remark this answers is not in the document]
or-else: Sure, just add shims in all adapters where it's not implemented.

server/db/mysql/adapter.go
or-else: How about just pulling the vars out of `Config`:
```go
type configType struct {
	ms.Config
        // not needed anymore: DBName string `json:"database,omitempty"`
        // Connection pool params.
	MaxOpenConns           int    `json:"max_open_conns,omitempty"`
	MaxIdleConns           int    `json:"max_idle_conns,omitempty"`
	ConnMaxLifetimeSeconds int    `json:"conn_max_lifetime_seconds,omitempty"`
}
```

aforge: Done. I've deprecated `dsn` and `database` fields.

server/db/mysql/adapter.go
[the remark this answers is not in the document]
aforge: Done.

server/db/mysql/adapter.go
```diff
@@ -42,11 +43,19 @@ const (
...
 type configType struct {
-	DSN    string `json:"dsn,omitempty"`
-	DBName string `json:"database,omitempty"`
+	ms.Config
+	// Connection pool settings.
+	MaxOpenConns           int `json:"max_open_conns,omitempty"`
+	MaxIdleConns           int `json:"max_idle_conns,omitempty"`
+	ConnMaxLifetimeSeconds int `json:"conn_max_lifetime_seconds,omitempty"`
+	// Deprecated.
+	DSN      string `json:"dsn,omitempty"`
```
or-else, line 57: You are hard-failing the initialization when the `DSN` is present (line 82 below). If so, why keep the DSN in the config at all? 

But I would give users a grace period: if `Config` is specified, ignore the `DSN`. Otherwise we are going to get a lot of complaints.

server/db/mysql/adapter.go
```diff
@@ -65,16 +74,21 @@ func (a *adapter) Open(jsonconfig json.RawMessage) error {
...
+	if config.DBName == "" {
+		config.DBName = defaultDatabase
 	}
-
 	a.dbName = config.DBName
-	if a.dbName == "" {
-		a.dbName = defaultDatabase
+
+	if config.DSN != "" {
+		return errors.New(fmt.Sprintf(mysqlConfigErrorText, "dsn"))
```
or-else, line 83: `errors.New(fmt.Sprintf(...))` -> `fmt.Errorf(...)`

I don't see `%s` in `mysqlConfigErrorText`, so `errors.New(mysqlConfigErrorText)` would produce the same result.

server/tinode.conf
```diff
@@ -190,7 +190,14 @@
 				// See https://github.com/go-sql-driver/mysql#dsn-data-source-name for syntax.
 				"dsn": "root@tcp(localhost)/tinode?parseTime=true&collation=utf8mb4_unicode_ci",
```
or-else, line 191: You are going to hard-fail on this config. Please provide a default working config.

server/tinode.conf
```diff
@@ -190,7 +190,14 @@
...
 				// Name of the main database.
-				"database": "tinode"
+				"database": "tinode",
```
or-else, line 193: Ditto here: the adapter returns an error when `"database"` is specified.

server/tinode.conf
```diff
@@ -190,7 +190,14 @@
...
+				// Maximum number of open connections to the database. Default: 0 (unlimited).
+				// "max_open_conns": 64,
```
or-else, line 195: Maybe not comment them out? There are reasonable values.

server/db/mysql/adapter.go
or-else: Or fail only if both config and DSN are specified.

server/db/mysql/adapter.go
```diff
@@ -65,16 +74,21 @@ func (a *adapter) Open(jsonconfig json.RawMessage) error {
...
+	}
+
+	if config.Database != "" {
+		return errors.New(fmt.Sprintf(mysqlConfigErrorText, "database"))
 	}
 
+	a.dsn = config.FormatDSN()
```
or-else, line 90: Forgot to mention. The driver's docs say:
```
NewConfig creates a new Config and sets default values.
```
It indeed sets some default values. We cannot use the `NewConfig` so the default values are not set in our case. Please check that the DSN created from our config is not affected by these unset default values. Thanks.

server/db/mysql/adapter.go
aforge: Done.

server/db/mysql/adapter.go
aforge: Yeah, that's a good observation. I am now using `NewConfig` to set the default values in `mysql.Config`.

server/tinode.conf
aforge: Done.

server/tinode.conf
aforge: SG. Done.

server/tinode.conf
aforge: Done.

aforge (commented):
Okay, I've modified the logic a little bit. Please take a look.

server/db/mysql/adapter.go
```diff
@@ -60,17 +72,33 @@ func (a *adapter) Open(jsonconfig json.RawMessage) error {
...
 		return errors.New("mysql adapter failed to parse config: " + err.Error())
 	}
 
-	a.dsn = config.DSN
-	if a.dsn == "" {
-		a.dsn = defaultDSN
+	locationCmp := cmp.Comparer(func(x, y time.Location) bool {
+		return x.String() == y.String()
+	})
+	if !cmp.Equal(*defaultCfg, config.Config, locationCmp, cmpopts.IgnoreUnexported(ms.Config{})) {
```
or-else, line 84: I don't understand this. `config.Config` is identical to `*defaultCfg`:
```go
config := configType{Config: *defaultCfg}
```
How can they ever be different?

or-else: I think it might be easier to just call `config.FormatDSN()` and compare it to the DSN you get from the default config. If they are different then Config must have been provided. If they are the same then the Config is blank.

or-else (approved):

server/db/mysql/adapter.go
aforge: That sounds reasonable. Changed.

PR #627, opened 2021-03-23
Stats() for rethinkdb and mongodb

or-else:


server/db/mongodb/adapter.go
```diff
@@ -233,7 +233,15 @@ func (a *adapter) Version() int {
 
 // DB connection stats object.
 func (a *adapter) Stats() interface{} {
-	return nil
+	if a.db == nil {
+		return nil
+	}
+
+	var result b.M
+	if err := a.db.RunCommand(a.ctx, b.D{{"connPoolStats", 1}}, nil).Decode(&result); err != nil {
```
aforge, line 241: I'm not sure if this is what we want. It always returns zeros for the local single replica setup.

https://docs.mongodb.com/manual/reference/command/connPoolStats/#definition
It says it returns connection stats between database instances rather than client -> server connection stats.

I think this should do better:
https://docs.mongodb.com/manual/reference/command/serverStatus/#connections

or-else: Changed to `serverStatus.connections`.
I saw the 0s but did not realise it was alway so.

aforge (approved):

PR #628, opened 2021-03-24
Fix docker readme environment variables table

andreas-eberle:
Currently the [table listing the available environment variables in the docker readme](https://github.com/tinode/chat/blob/master/docker/README.md#supported-environment-variables) is not correctly shown. This is caused by an additional newline which breaks the markdown table formatting. This simple PR removes the newline so the table is redered correctly.


or-else:
Thanks!

PR #630, opened 2021-03-29
Mysql: query timeouts.

aforge:
Configure and enforce timeouts on DB queries.
This way pending queries won't have to wait and queue indefinitely long until
connection pool resources become available.

### DIFF
--- a/server/db/mysql/adapter.go
+++ b/server/db/mysql/adapter.go
@@ -64,9 +64,13 @@ type configType struct {
 	Database string `json:"database,omitempty"`
 
 	// Connection pool settings.
-	MaxOpenConns           int `json:"max_open_conns,omitempty"`
-	MaxIdleConns           int `json:"max_idle_conns,omitempty"`
-	ConnMaxLifetimeSeconds int `json:"conn_max_lifetime_seconds,omitempty"`
+	//
+	// Maximum number of open connections to the database.
+	MaxOpenConns int `json:"max_open_conns,omitempty"`
+	// Maximum number of connections in the idle connection pool.
+	MaxIdleConns int `json:"max_idle_conns,omitempty"`
+	// Maximum amount of time a connection may be reused (in seconds).
+	ConnMaxLifetime int `json:"conn_max_lifetime,omitempty"`
 
 	// DB request timeout (in seconds).
 	// If 0 (or negative), no timeout is applied.
@@ -154,8 +158,8 @@ func (a *adapter) Open(jsonconfig json.RawMessage) error {
 		if config.MaxIdleConns > 0 {
 			a.db.SetMaxIdleConns(config.MaxIdleConns)
 		}
-		if config.ConnMaxLifetimeSeconds > 0 {
-			a.db.SetConnMaxLifetime(time.Duration(config.ConnMaxLifetimeSeconds) * time.Second)
+		if config.ConnMaxLifetime > 0 {
+			a.db.SetConnMaxLifetime(time.Duration(config.ConnMaxLifetime) * time.Second)
 		}
 		if config.SqlTimeout > 0 {
 			a.sqlTimeout = time.Duration(config.SqlTimeout) * time.Second
--- a/server/db/mysql/adapter.go
+++ b/server/db/mysql/adapter.go
@@ -31,8 +31,9 @@ type adapter struct {
 	maxMessageResults int
 	version           int
 
-	// Single query and transaction timeouts (in seconds).
+	// Single query timeout.
 	sqlTimeout time.Duration
+	// DB transaction timeout.
 	txTimeout  time.Duration
 }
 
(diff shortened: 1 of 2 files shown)


server/db/mysql/adapter.go
```diff
@@ -29,6 +30,10 @@ type adapter struct {
 	// Maximum number of message records to return
 	maxMessageResults int
 	version           int
+
+	// Single query and transaction timeouts (in seconds).
```
or-else, line 34: Go tools like IDEs show comments in tool tips when the mouse hovers over a variable. It's a good practice to comment each variable or function individually.

server/db/mysql/adapter.go
```diff
@@ -57,6 +62,25 @@ type configType struct {
 	MaxOpenConns           int `json:"max_open_conns,omitempty"`
 	MaxIdleConns           int `json:"max_idle_conns,omitempty"`
 	ConnMaxLifetimeSeconds int `json:"conn_max_lifetime_seconds,omitempty"`
+
+	// Single query and transaction timeouts in seconds.
+	// If 0 (or negative), no timeout is applied.
+	SqlTimeoutSeconds int `json:"sql_timeout_seconds,omitempty"`
```
or-else, line 68: The `_seconds` in `sql_timeout_seconds` is good and informative but inconsistent with how it was done before. It implies it was not seconds elsewhere or possibly creates questions. Maybe drop the `_seconds` from the variable name and leave it in comments only?

server/db/mysql/adapter.go
```diff
@@ -156,8 +186,12 @@ func (a *adapter) GetDbVersion() (int, error) {
 		return a.version, nil
 	}
 
+	ctx, cancel := a.getContext()
```
or-else, line 189: The purpose of `cancel` is to cancel the query asynchronously, i.e. to launch the query from one go routine then cancel it from another. I don't think we will ever make such async DB calls consequently `cancel` is unlikely to be used ever. If it does not do anything now and unlikely to be used in the future it would make sense to get rid of it completely and remove all the associated boilerplate code.

This combined with the single timeout I mentioned above would allow calls like `a.db.GetContext(a.ctx(), ...)`.

server/db/mysql/adapter.go
```diff
@@ -57,6 +62,25 @@ type configType struct {
 	MaxOpenConns           int `json:"max_open_conns,omitempty"`
 	MaxIdleConns           int `json:"max_idle_conns,omitempty"`
 	ConnMaxLifetimeSeconds int `json:"conn_max_lifetime_seconds,omitempty"`
+
+	// Single query and transaction timeouts in seconds.
+	// If 0 (or negative), no timeout is applied.
+	SqlTimeoutSeconds int `json:"sql_timeout_seconds,omitempty"`
+	TxTimeoutSeconds  int `json:"tx_timeout_seconds,omitempty"`
```
or-else, line 69: Should we have two separate timeouts?

server/db/mysql/adapter.go
```diff
@@ -2366,7 +2573,7 @@ func (a *adapter) MessageAttachments(msgId t.Uid, fids []string) error {
 		}
 	}()
 
-	_, err = a.db.Exec("INSERT INTO filemsglinks(createdat,fileid,msgid) VALUES "+strings.Join(values, ","), args...)
+	_, err = tx.Exec("INSERT INTO filemsglinks(createdat,fileid,msgid) VALUES "+strings.Join(values, ","), args...)
```
or-else, line 2576: Nice catch!

server/db/mysql/adapter.go
aforge: Done.

server/db/mysql/adapter.go
aforge: Done.

server/db/mysql/adapter.go
or-else: OK, so it's an optimization which allows to free resources explicitly sooner, without waiting for the GC. OK.

server/db/mysql/adapter.go
```diff
@@ -57,6 +63,27 @@ type configType struct {
 	MaxOpenConns           int `json:"max_open_conns,omitempty"`
 	MaxIdleConns           int `json:"max_idle_conns,omitempty"`
 	ConnMaxLifetimeSeconds int `json:"conn_max_lifetime_seconds,omitempty"`
```
or-else, line 65: And maybe this one too for consistency.

or-else (approved):

server/db/mysql/adapter.go
aforge: Done.

server/db/mysql/adapter.go
aforge: Done.

Issue #631, opened 2021-03-30
Mysql adapter optimisations


gabriel-vasile:
I understand [people have different levels of experience](https://github.com/tinode/chat/pull/609#issuecomment-765684294), but could you, please, NOT optimise sections like this? https://github.com/tinode/chat/blob/e64742f3f0c1f5c3e62dfc86703d086c288588d2/server/db/mysql/adapter.go#L827-L842

or-else:
Could you please clarity what exactly is the problem with this code?

gabriel-vasile:
It reuses the same user var to save some allocs.
Fortunately, builtin `append` func is not so concerned about allocs and it makes a copy of the user when called so most of user data is correct.
Unfortunately, there are some pass by reference vars inside user. 

This example uses JSON, but the problem reproduces the same with any decoding func, including `sqlx`'s `StructScan`.
https://play.golang.org/p/FBndxEXzUeZ

Unless the code is in a tight loop with thousands of executions per second, there is no reason for such optimizations. Most of the times memory goes on stack which is insanely fast. Plus, it's quite hard to say there is an improvement when there are no benchmarks to prove it.

or-else:
Good catch with slices. Thanks!

Issue #604, opened 2021-01-18
Cannot get latest group topic


wmshek1992:
### Cannot get latest group topic
One of my user has 1085 topics, most of them are group topics and a few p2p topics.
I tried to use {sub: {ims: "2021-01-06T18:07:30.038Z"}} to limit the result.
I expect the result would only consist of topics that their "updated" are after "ims", but it failed to work. 
So what is the correct way to obtain a response with latest topics

#### Server-side
  * version of Tinode server 0.16.4
  * database backend : MySQL
  * standalone

or-else:
Yes, it's a design limitation. In the short term you can change this line and recompile:
https://github.com/tinode/chat/blob/master/server/db/mysql/adapter.go#L42
I'll fix it in the next release.

or-else:
Actually, you don't need to recompile. Set max_results in store config:
https://github.com/tinode/chat/blob/master/server/tinode.conf#L174

Set it to something like 2000 and restart the server

wmshek1992:
Yes, I increased the limit in the config file. But potentially it could be exceeded in long term, and also the result can be huge. 

I would like to limit the topic list result by it's updated time such that clients can get the latest topics. Is it an existing feature or a new feature request?

or-else:
As I said in my previous message, I'll fix it in 0.17

or-else:
The same as #342

or-else:
The `ims` in `{get what=subs}` is now respected. The clients are not updated to use it but the server will paginate the subscriptions by `ims`.

or-else:
I'm closing this, please continue further discussion in #342.

or-else:
Duplicate of #342.

PR #633, opened 2021-04-01
Clean up deletions in mongodb adapter

or-else:
* Soft-deleted message records were not properly cleaned up on user and subscription deletions. The same fix as in rethink & mysql.
* Bug fixed in standalone mongo when deletions were failing with '(IllegalOperation) Transaction numbers are only allowed on a replica set member or mongos' due to `RetriableWrites` being `true` by default.
* Fix for #631 (reused slices)
* Removed unused parameters from adapter.SubsForUser (first step for #604)

### DIFF
--- a/server/db/mongodb/adapter.go
+++ b/server/db/mongodb/adapter.go
@@ -1753,7 +1753,7 @@ func (a *adapter) SubsDelete(topic string, user t.Uid) error {
 
 	forUser := user.String()
 
-	if err = mdb.WithSession(a.ctx, sess, func(sc mdb.SessionContext) error {
+	return mdb.WithSession(a.ctx, sess, func(sc mdb.SessionContext) error {
 		if err := a.subsDelete(sc, b.M{"_id": topic + ":" + forUser}, false); err != nil {
 			return err
 		}
@@ -1774,10 +1774,7 @@ func (a *adapter) SubsDelete(topic string, user t.Uid) error {
 		}
 		// Commit changes.
 		return a.maybeCommitTransaction(sc, sess)
-	}); err != nil {
-		return err
-	}
-	return nil
+	})
 }
 
 // Delete/mark deleted subscriptions.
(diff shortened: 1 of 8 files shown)


server/db/mongodb/adapter.go
```diff
@@ -1729,10 +1739,45 @@ func (a *adapter) SubsUpdate(topic string, user t.Uid, update map[string]interfa
...
+			filter := b.M{"topic": topic, "deletedfor.user": forUser}
+			if _, err := a.db.Collection("messages").
+				UpdateMany(sc, filter, b.M{"$pull": b.M{"deletedfor": b.M{"user": forUser}}}); err != nil {
+				return err
+			}
+		}
+		// Commit changes.
+		return a.maybeCommitTransaction(sc, sess)
+	}); err != nil {
+		return err
```
aforge, line 1778: Maybe just
```
return mdb.WithSession(a.ctx, sess, func(sc mdb.SessionContext) error {...})
```
?

or-else: done

server/db/mysql/adapter.go
```diff
@@ -1900,33 +1900,12 @@ func (a *adapter) SubscriptionGet(topic string, user t.Uid) (*t.Subscription, er
...
-
-		if opts.Topic != "" {
-			q += " AND topic=?"
-			args = append(args, opts.Topic)
-		}
-		if opts.Limit > 0 && opts.Limit < limit {
-			limit = opts.Limit
-		}
-	}
-	q += " LIMIT ?"
```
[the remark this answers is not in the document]
or-else: Yes, I am, but we need to find a way to fix it in another way. This limit here behaves like a bug: users with too many subscriptions stop receiving push and presence notifications for some subscriptions. We need to limit the number of subscriptions at the point where the user creates them.

aforge (approved):

PR #635, opened 2021-04-10
Update README.md of docker-specific instructions

remfath:
fix the table layout of environment variables


or-else:
Thanks. JFYI, it's been fixed already in devel.

PR #641, opened 2021-04-16
Add a basic unit test for topic.go.

aforge:
Steps:
1. Wrap store methods in interfaces (in order to allow method mocking).
2. Generate store/mock_store/mock_store.go with mockgen (mock objects).
3. Add a couple of basic tests for Topic.handleBroadcast.


or-else:
Great idea. 

Just a question: would it make more sense to create a mock DB adapter instead of mocking the Store?

aforge:
> Great idea.
> 
> Just a question: would it make more sense to create a mock DB adapter instead of mocking the Store?

Yeah, I tried that first. The problem is you'd have to initialize the adapter first, register it with the store and expose the internals of the `store` code/logic. This makes the code longer, less readable and harder to change. Here's an equivalent of mocking `store.Messages.Save`:
```go
        a := mock_adapter.NewMockAdapter(ctrl)

        a.EXPECT().GetName().Return("dummy")
        a.EXPECT().IsOpen().Return(false)
        a.EXPECT().Open(gomock.Any())
        a.EXPECT().SetMaxResults(gomock.Any())
        a.EXPECT().CheckDbVersion()
        a.EXPECT().TopicUpdateOnMessage("p2p-test", gomock.Any()).Return(nil)
        a.EXPECT().MessageSave(gomock.Any()).Return(nil)
        a.EXPECT().SubsUpdate("p2p-test", uid1, gomock.Any())

        store.RegisterAdapter(a)
        storeConfig := []byte(`{"use_config": "dummy",
                          "uid_key": "la6YsO+bNX/+XIkOqc5Svw==" }`)
        if err := store.Open(0, storeConfig); err != nil {
                log.Fatal("Failed to initialize data store:", err)
        }
```

or-else (approved):

PR #654, opened 2021-05-26
Make relative paths resolve against executable directory.

aforge:
To make things consistent with how the server treats paths.

### DIFF
--- a/server/main.go
+++ b/server/main.go
@@ -15,7 +15,6 @@ import (
 	"flag"
 	"net/http"
 	"os"
-	"path/filepath"
 	"runtime"
 	"runtime/pprof"
 	"strings"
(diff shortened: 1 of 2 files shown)


or-else:
I don't remember why I made it to resolve against the exec path. The resolution against the current dir looks more conventional. Maybe do it the other way around?

or-else:
Actually, I do remember now - that made some sense because of the way sandbox was deployed back then. I don't think it matters anymore. Current dir makes more sense.

aforge:
Changed it to use cwd for the config.
I'm keeping it executable path for the templates dir as this comment explains:
https://github.com/tinode/chat/blob/devel/server/main.go#L564-L571

Or do you think we should change it to the cwd as well?

or-else (approved):

or-else:
I would resolve all paths from the CWD. It's the option of least surprise.

PR #660, opened 2021-06-02
Send 'sub' push notifications when user joins a group.

aforge:
We need to notify group members of a new subscription.
Presently, this is only available in p2p topic.

This is the reason why new subscriptions show up as unknown (id) in group chats in certain cases.

### DIFF
--- a/server/topic.go
+++ b/server/topic.go
@@ -810,13 +810,16 @@ func (t *Topic) sendImmediateSubNotifications(asUid types.Uid, acs *MsgAccessMod
 			t.presSingleUserOffline(uid2, mode2, status, nilPresParams, "", false)
 
 			// Also send a push notification to the other user.
-			if pushRcpt := t.pushForSub(asUid, uid2, pud2.modeWant, pud2.modeGiven, types.TimeNow()); pushRcpt != nil {
+			if pushRcpt := t.pushForP2PSub(asUid, uid2, pud2.modeWant, pud2.modeGiven, types.TimeNow()); pushRcpt != nil {
 				usersPush(pushRcpt)
 			}
 		}
 	} else if t.cat == types.TopicCatGrp {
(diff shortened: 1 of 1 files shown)


server/topic.go
```diff
@@ -814,6 +814,10 @@ func (t *Topic) sendImmediateSubNotifications(asUid types.Uid, acs *MsgAccessMod
 				usersPush(pushRcpt)
 			}
 		}
+	} else if t.cat == types.TopicCatGrp {
+		if pushRcpt := t.pushForGroupSub(asUid, types.TimeNow()); pushRcpt != nil {
```
or-else, line 818: Please also check if `sreg.pkt.Sub.Newsub` is `true`. Otherwise pushes will be sent on all permission changes.

server/topic.go
```diff
@@ -3234,14 +3237,39 @@ func (t *Topic) pushForSub(fromUid, toUid types.Uid, want, given types.AccessMod
...
 			SeqId:     t.lastID,
-			ModeWant:  want,
-			ModeGiven: given,
 		},
 	}
+	return receipt
+}
+
+// Prepares payload to be delivered to a mobile device as a push notification in response to a new subscription in a p2p topic.
+func (t *Topic) pushForSub(fromUid, toUid types.Uid, want, given types.AccessMode, now time.Time) *push.Receipt {
```
or-else, line 3246: Rename to `pushForP2PSub`?

or-else (approved):
There was a bug in Android which also independently caused 'Not found' group members:
https://github.com/tinode/tindroid/commit/055f8b5db000d26fed1b4e55890f82a14b83f67e
Specifically changes like this fixed it:
```java
if (isP2PType()) {
    mTinode.updateUser(getName(), mDesc);
}
```
It probably exists in iOS too.

server/topic.go

server/topic.go

PR #662, opened 2021-06-12
Updated TopicsForUser in rethinkdb, refactored common code, added tests

or-else:
* Fixed a few minor issues in TopicsForUser in MySQL.
* Updated TopicsForUser in RethinkDB to use the same logic as MySQL
* Moved common code to a separate package
* Added tests

Mongo is going to be next.

Update: added mongo, fixed some bugs, updated mongo tests.

### DIFF
--- a/server/db/common/common.go
+++ b/server/db/common/common.go
@@ -1,0 +1,63 @@
+// Package common contains utility methods used by all adapters.
+package common
+
+import (
+	"sort"
+	"time"
+
+	t "github.com/tinode/chat/server/store/types"
+)
+
+// SelectEarliestUpdatedSubs selects no more than the given umber of subscriptions from the
+// given slice satisfying the query. When the number of subscriptions is greater than the limit,
+// the subscriptions with the earliest timestamp are selected.
+func SelectEarliestUpdatedSubs(subs []t.Subscription, opts *t.QueryOpt, maxResults int) []t.Subscription {
+	limit := maxResults
+	ims := time.Time{}
+	if opts != nil {
+		if opts.Limit > 0 && opts.Limit < limit {
+			limit = opts.Limit
+		}
+		if opts.IfModifiedSince != nil {
+			ims = *opts.IfModifiedSince
+		}
+	}
+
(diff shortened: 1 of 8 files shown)


server/db/common/common.go
```diff
@@ -0,0 +1,63 @@
...
+package common
+
+import (
+	"sort"
+	"time"
+
+	t "github.com/tinode/chat/server/store/types"
+)
+
+// SelectEarliestUpdatedSubs selects no more than the given umber of subscriptions from the
```
aforge, line 11: s/he given umber of subscriptions/`numResults`

server/db/common/common.go
```diff
@@ -0,0 +1,63 @@
...
+// given slice satisfying the query. When the number of subscriptions is greater than the limit,
+// the subscriptions with the earliest timestamp are selected.
+func SelectEarliestUpdatedSubs(subs []t.Subscription, opts *t.QueryOpt, maxResults int) []t.Subscription {
```
aforge, line 14: Actually thinking a bit more about it:
do we want more stale (i.e. updated at an earlier time) subs?
Should we return the latest updated subs instead (i.e. updated at a later time)?

aforge (approved):

server/db/common/common.go
or-else: done

server/db/common/common.go
or-else: Yes, we do:

1. Client has subscriptions with the latest update timestamp X.
2. Client asks for N updated subscriptions since X. The server returns N with updates between X and Y. 
3. Go to 1 with X := Y.

If the server returns the the most recent N subscriptions, then there could be a gap between X and the oldest timestamp in the second batch. 

It's a different logic than iterating messages because gaps are acceptable for messages. The gaps are not acceptable for subscriptions.

aforge: I see. Thanks for clarifying. Perhaps, add it as a comment in the code?

or-else: done

PR #684, opened 2021-07-21
Update API.md

psvnlsaikumar:
Added an external wiki link to pub-sub design. Thought it would be helpful for anyone for a quick reference instead of a google search.


or-else:
Thanks.

PR #699, opened 2021-09-08
Adding support for multiple smtp auth mechanism

rkgarcia:
We are adding this to use Office 365 account for notifications

### DIFF
--- a/server/tinode.conf
+++ b/server/tinode.conf
@@ -325,6 +325,9 @@
 				// Password to use when authenticating the sender; used only if "login" is provided.
 				"sender_password": "your-password-here",
 
+				// Mechanism supported for authentication we can use plain, login or md5
+				"mechanism": "login",
+
 				// FQDN to use in SMTP HELO/EHLO command; if missing, the hostname from "host_url" is used.
 				"smtp_helo_host": "example.com",
 
--- a/server/tinode.conf
+++ b/server/tinode.conf
@@ -325,8 +325,8 @@
 				// Password to use when authenticating the sender; used only if "login" is provided.
 				"sender_password": "your-password-here",
 
-				// Mechanism supported for authentication we can use plain, login or md5
-				"mechanism": "login",
+				// Authentication mechanism to use, optional. One of "login", "md5", "plain" (default).
+				"auth_mechanism": "login",
 
 				// FQDN to use in SMTP HELO/EHLO command; if missing, the hostname from "host_url" is used.
 				"smtp_helo_host": "example.com",
--- a/server/validate/email/validate.go
+++ b/server/validate/email/validate.go
@@ -41,6 +41,8 @@ type validator struct {
 	Login string `json:"login"`
 	// Password to use for SMTP authentication.
 	SenderPassword string `json:"sender_password"`
+	// Optional to use for PLAIN, LOGIN or MD5 mechanism
+	Mechanism string `json:"mechanism"`
 	// Optional response which bypasses the validation.
 	DebugResponse string `json:"debug_response"`
 	// Number of validation attempts before email is locked.
@@ -119,6 +121,31 @@ func isTemplateValid(templ *textt.Template) error {
 	}
 	return nil
 }
+type SMPTLogin struct {
+	username, password string
+}
+
+func LoginAuth(username, password string) smtp.Auth {
+	return &SMPTLogin{username, password}
+}
+
+func (a *SMPTLogin) Start(server *smtp.ServerInfo) (string, []byte, error) {
+	return "LOGIN", []byte(a.username), nil
+}
+
+func (a *SMPTLogin) Next(fromServer []byte, more bool) ([]byte, error) {
+	if more {
+		switch string(fromServer) {
+		case "Username:":
+			return []byte(a.username), nil
+		case "Password:":
+			return []byte(a.password), nil
+		default:
+			return nil, errors.New("Unknown from server")
+		}
+	}
+	return nil, nil
+}
 
 type emailContent struct {
 	subject string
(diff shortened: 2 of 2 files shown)


server/validate/email/validate.go
```diff
@@ -41,6 +41,8 @@ type validator struct {
 	Login string `json:"login"`
 	// Password to use for SMTP authentication.
 	SenderPassword string `json:"sender_password"`
+	// Optional to use for PLAIN, LOGIN or MD5 mechanism
+	Mechanism string `json:"mechanism"`
```
or-else, line 45: Please rename it to 
```go
AuthMechanism string `json:"auth_mechanism"`
```

Otherwise it's a bit ambigous.

server/validate/email/validate.go
```diff
@@ -41,6 +41,8 @@ type validator struct {
 	Login string `json:"login"`
 	// Password to use for SMTP authentication.
 	SenderPassword string `json:"sender_password"`
+	// Optional to use for PLAIN, LOGIN or MD5 mechanism
```
or-else, line 44: Please mention that the default is PLAIN, something like 
`// Authentication mechanism to use, optional. One of "login", "md5", "plain" (default).`

server/tinode.conf
```diff
@@ -325,6 +325,9 @@
 				// Password to use when authenticating the sender; used only if "login" is provided.
 				"sender_password": "your-password-here",
 
+				// Mechanism supported for authentication we can use plain, login or md5
```
or-else, line 328: `// Authentication mechanism to use, optional. One of "login", "md5", "plain" (default).`

server/tinode.conf
```diff
@@ -325,6 +325,9 @@
...
+				"mechanism": "login",
```
or-else, line 329: Please rename to `auth_mechanism`, otherwise it is ambiguous without the comment.

server/validate/email/validate.go
```diff
@@ -119,6 +121,31 @@ func isTemplateValid(templ *textt.Template) error {
 	}
 	return nil
 }
+type SMPTLogin struct {
```
or-else, line 124: `SMPTLogin`  -> `SMTPLogin`

or-else: I don't think you need to export this structure from the module. Consequently, please change the name to have a lowercase first letter. 
If you believe it needs to be exported, please explain why and add a required godoc comment.

I would name it `loginAuth` to follow the same pattern as `plainAuth`.

server/validate/email/validate.go
```diff
@@ -119,6 +121,31 @@ func isTemplateValid(templ *textt.Template) error {
...
+}
+
+func (a *SMPTLogin) Start(server *smtp.ServerInfo) (string, []byte, error) {
+	return "LOGIN", []byte(a.username), nil
+}
+
+func (a *SMPTLogin) Next(fromServer []byte, more bool) ([]byte, error) {
+	if more {
+		switch string(fromServer) {
+		case "Username:":
```
or-else, line 139: Are you sure it's always capitalized? Maybe apply `ToLower`?

server/validate/email/validate.go
```diff
@@ -119,6 +121,31 @@ func isTemplateValid(templ *textt.Template) error {
...
+			return []byte(a.username), nil
+		case "Password:":
```
or-else, line 141: Are you sure it's always capitalized? Maybe apply `ToLower`?

server/validate/email/validate.go
```diff
@@ -119,6 +121,31 @@ func isTemplateValid(templ *textt.Template) error {
...
+			return []byte(a.password), nil
+		default:
+			return nil, errors.New("Unknown from server")
```
or-else, line 144: Maybe make the error more descriptive, like 'Unknown server response in LOGIN AUTH' or something like that.

server/validate/email/validate.go
```diff
@@ -119,6 +121,31 @@ func isTemplateValid(templ *textt.Template) error {
 	}
 	return nil
 }
+type SMPTLogin struct {
+	username, password string
+}
+
+func LoginAuth(username, password string) smtp.Auth {
```
or-else, line 128: I don't think you need to export this method from the module. Consequently, please change the name to have a lowercase first letter. If you believe it needs to be exported, please explain why and add a required godoc comment.

or-else (changes requested):

server/validate/email/validate.go
or-else: Actually, you don't even need this method. You can initialize the struct directly. It's only needed if you are writing a module, but you are not.

server/validate/email/validate.go
```diff
@@ -169,7 +196,15 @@ func (v *validator) Init(jsonconf string) error {
 
 	// Enable auth if login is provided.
 	if v.Login != "" {
-		v.auth = smtp.PlainAuth("", v.Login, v.SenderPassword, v.SMTPAddr)
+		mechanism := strings.ToLower(v.Mechanism)
+		switch mechanism {
+		case "md5":
+			v.auth = smtp.CRAMMD5Auth(v.Login, v.SenderPassword)
+		case "login":
+			v.auth = LoginAuth(v.Login, v.SenderPassword)
```
or-else, line 204: You can do it directly here:
```
v.auth = &loginAuth{v.Login, v.SenderPassword}
```

server/validate/email/validate.go
```diff
@@ -169,7 +196,15 @@ func (v *validator) Init(jsonconf string) error {
...
+		default:
```
or-else, line 205: Actually, this is not a good idea as it would lead to strange behavior. Please change it to 
```go
case "", "plain":
    v.auth = smtp.PlainAuth("", v.Login, v.SenderPassword, v.SMTPAddr)
default:
    ... fail with an error ...
```

rkgarcia: I will change with

```go
case "", "plain":
	v.auth = smtp.PlainAuth("", v.Login, v.SenderPassword, v.SMTPAddr)
default:
	return errors.New("Bad Auth Mechanism for SMTP")
```

or-else (approved):

PR #700, opened 2021-09-08
Drafty handling refactored to use trees

or-else:
Drafty formatting refactored to use trees: a doc is converted to a tree where each node is uniformly formatted, like

[_quick **fox** jumps_ over] ->
```
[[EM: quick, [ST: fox], jumps], over]
```
It simplified removal of quoted text, shortening etc. Also added a few more tests for various corner cases.

--

There are three unrelated changes:
* `auth_basic.go`: added '**.**' as a permitted character to logins: https://groups.google.com/g/tinode/c/Ncl0MG-QzQM/m/cvi88V0ACAAJ
* Added a comment to mysql `schema.sql` as a reaction to this: https://groups.google.com/g/tinode/c/meEqSq6XzKU/m/tzGrao6-CAAJ
* Fixed a bug in `store.go` related to handling of out-of-band uploaded files.

### DIFF
--- a/server/db/mysql/schema.sql
+++ b/server/db/mysql/schema.sql
@@ -1,3 +1,13 @@
+# THIS SCHEMA FILE IS FOR REFERENCE/DOCUMENTATION ONLY!
+# DO NOT USE IT TO INITIALIZE THE DATABASE.
+# Read installation instructions first.
+
+# The following line will produce an intentional error.
+
+'READ INSTALLATION INSTRUCTIONS!';
+
+# The actual schema is below.
+
 DROP DATABASE IF EXISTS tinode;
 
 CREATE DATABASE tinode CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci;
--- a/server/drafty/drafty.go
+++ b/server/drafty/drafty.go
@@ -53,36 +53,6 @@ type node struct {
 	children []*node
 }
 
-func (n *node) String() string {
-	str := "{"
-	if n == nil {
-		str = "nil"
-	} else {
-		if n.sp != nil {
-			str += n.sp.tp
-			if n.sp.data != nil {
-				str += ":data"
-			}
-			str += ":"
-		}
-
-		if n.txt != nil {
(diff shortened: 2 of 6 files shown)


server/db/mysql/schema.sql
```diff
@@ -1,3 +1,13 @@
+# THIS SCHEMA FILE IS FOR REFERENCE/DOCUMENTATION ONLY!
+# DO NOT USE IT TO INITIALIZE THE DATABASE.
+# Read installation instructions first.
+
+# The following line will produce an intentional error.
+
+'READ INSTALLATION INSTRUCTIONS!';
```
aforge, line 7: Perhaps, add a link here?

aforge (commented):
How hard would it be to add a test for `drafty.go`?

or-else:
> How hard would it be to add a test for drafty.go?

Which test do you mean? There are a few unit tests there.

server/db/mysql/schema.sql
or-else: I thought about it and decided against it: we will forget about it and it will get stale. It's just a forcing function to make people search for answers.

server/drafty/drafty.go
```diff
@@ -22,368 +47,572 @@ type span struct {
...
-	"CO": {"", false},
+	"CO": {"`", false},
 	"BR": {"\n", true},
 	"LN": {"", false},
 	"MN": {"", false},
 	"HT": {"", false},
-	"IM": {"", true},
+	"IM": {"", false},
 	"EX": {"", true},
+	"QQ": {"", false},
```
aforge, line 168: Can you add `IC` here as well?

aforge:
> > How hard would it be to add a test for drafty.go?
> 
> Which test do you mean? There are a few unit tests there.

My bad. Sorry for confusion.

aforge (approved):

server/drafty/drafty.go
or-else: Actually, only non-blank tags are needed here. I added `IC` handling further down in the code + test.

PR #710, opened 2021-11-07
Fixing {get sub ims=...} for subscriptions with new messages

or-else:
See context here:
https://groups.google.com/g/tinode/c/nyXwZvff6SM/m/HFoGeqOZBQAJ

If the topic received a new message while the client was offline with no other changes, the next {sub topic=me ims=...} returned an empty set effectively not reporting the new message. The problem was only when the device had pushes disabled.


server/db/common/common.go
```diff
@@ -23,41 +23,39 @@ func SelectEarliestUpdatedSubs(subs []t.Subscription, opts *t.QueryOpt, maxResul
...
-			return subs[i].UpdatedAt.Before(subs[j].UpdatedAt)
+	// No cache management and the number of results is below the limit: return all.
+	if ims.IsZero() && len(subs) <= limit {
+		return subs
+	}
+
+	// Now that we fetched potentially more subscriptions than needed, we got to take those with the oldest modifications.
+	// Sorting in ascending order by modification time.
+	sort.Slice(subs, func(i, j int) bool {
+		return SelectLatestTime(subs[i].UpdatedAt, subs[i].GetTouchedAt()).
```
aforge, line 34: Perhaps for better readbility, make a helper function for
`SelectLatestTime(s.UpdatedAt, s.GetTouchedAt())`, e.g. `LastModified`.

aforge (approved):

PR #712, opened 2021-11-15
mysql-adapter: fix deadlock in SubsDelete if record is not found

bkleiner:
`err` being nil causes the rollback to be skipped


or-else:
Thanks!

PR #736, opened 2022-03-24
translate to vietnamese

ducmami:


server/templ/email-password-reset-vi.templ
```diff
@@ -0,0 +1,62 @@
...
+  ENGLISH
+
+  This template defines contents of the password reset email.
+
+  See explanation in ./email-validation-en.templ
+*/}}
+
+
+{{define "subject" -}}
+Reset Tinode password
```
or-else, line 11: This line needs to be translated too.

or-else (commented):
Thanks!

PR #760, opened 2022-05-29
Video call state moved to headers

aforge:
Call state/duration moved to replacement msg headers.
    
'webrtc' header contains state (started, accepted, finished, disconnected).
'webrtc-duration' header is for the server-computed call duration.


server/calls.go
```diff
@@ -52,12 +52,20 @@ type videoCall struct {
...
 	// Call message seq ID.
 	seq int
+	// Call message content.
+	content interface{}
+	// Time when the call was accepted.
+	acceptedAt time.Time
 }
 
-func (call *videoCall) messageHead() map[string]interface{} {
+func (call *videoCall) messageHead(newState string, duration int64) map[string]interface{} {
```
or-else, line 61: `int64` looks like an overkill. `int` should be plenty enough.

server/calls.go
```diff
@@ -52,12 +52,20 @@ type videoCall struct {
...
 	head := make(map[string]interface{})
 	head["mime"] = constTinodeVideoCallMimeType
```
or-else, line 63: Remove? Change to Drafty mime?

or-else (approved):

server/calls.go
aforge: Changed.

server/calls.go
aforge: Removed.

PR #779, opened 2022-08-03
TNPG updated to work with FCM push v1.

or-else:
TNPG updated
FCM push refactored a bit to use the new error handling. 
Common code moved to a separate package.

All pushes seem to work, including voip. But voip pushes arrive as normal pushes as opposite to PushKit despite the fact that they are sent as voip.

### DIFF
--- a/server/push/fcm/payload.go
+++ b/server/push/fcm/payload.go
@@ -93,11 +93,11 @@ func clonePayload(src map[string]string) map[string]string {
 
 // PrepareV1Notifications creates notification payloads ready to be posted
 // to push notification server for the provided receipt.
-func PrepareV1Notifications(rcpt *push.Receipt, config *configType) []*fcmv1.Message {
+func PrepareV1Notifications(rcpt *push.Receipt, config *configType) ([]*fcmv1.Message, []t.Uid) {
 	data, err := payloadToData(&rcpt.Payload)
 	if err != nil {
 		logs.Warn.Println("fcm push: could not parse payload:", err)
-		return nil
+		return nil, nil
 	}
 
 	// Device IDs to send pushes to.
@@ -122,11 +122,11 @@ func PrepareV1Notifications(rcpt *push.Receipt, config *configType) []*fcmv1.Mes
 		devices, count, err = store.Devices.GetAll(uids...)
 		if err != nil {
 			logs.Warn.Println("fcm push: db error", err)
-			return nil
+			return nil, nil
 		}
 	}
 	if count == 0 && rcpt.Channel == "" {
-		return nil
+		return nil, nil
 	}
 
 	if config == nil {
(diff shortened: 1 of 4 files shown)


or-else:
The new PushGW is also in `fcmv1` branch. It needs work though: I converted it to use v1 only, but it still needs to support v0 while we transition. i'll add v0 back.

server/push/fcm/payload.go
```diff
@@ -597,10 +196,11 @@ func PrepareV1Notifications(rcpt *push.Receipt, config *configType) []*fcmv1.Mes
 		// We don't know the platform of the receiver, must provide payload for all platforms.
 		msg.Android = androidNotificationConfig(rcpt.Payload.What, topic, userData, config)
 		msg.Apns = apnsNotificationConfig(rcpt.Payload.What, topic, userData, 0, config)
+		// TODO: add webpush payload.
 		messages = append(messages, &msg)
 	}
 
-	return messages
+	return messages, uids
```
aforge, line 203: Is `len(messages) == len(uids)` an invariant here?
If it is, channel messages (line 200) breaks it.

server/push/tnpg/push_tnpg.go
```diff
@@ -251,30 +277,35 @@ func processSubscription(req *push.ChannelReq, config *configType) {
...
 
-func handlePushResponse(batch *batchResponse, messages []fcm.MessageData) {
+func handlePushResponse(batch *batchResponse, messages []*fcmv1.Message, uids []types.Uid) {
 	if batch.FailureCount <= 0 {
 		return
 	}
 
 	for i, resp := range batch.Responses {
+
+		logs.Warn.Println("tnpg error:", resp.ErrorCode, resp.ErrorMessage, resp.ExtendedError, resp.Code)
```
aforge, line 287: remove?

aforge:
> But voip pushes arrive as normal pushes as opposite to PushKit despite the fact that they are sent as voip.

Humm... that sounds like FCM won't actually send PushKit/voip pushes?
It defeats the purpose of this change...

server/push/tnpg/push_tnpg.go
or-else: Removed.

server/push/fcm/payload.go
or-else: It should although the difference does not affect anything. Fixed anyway.

or-else:
> It defeats the purpose of this change...

Some people say it should work, some that it won't. FCM returns no error when I send voip pushes. I'll try to debug it first. Adding APNS is a lot of work.

or-else:
Alternatively maybe we can use something like https://github.com/appleboy/gorush

aforge (approved):

PR #781, opened 2022-08-08
delete minor unreachable code caused by t.Fatal

Abirdcfly:
Signed-off-by: Abirdcfly <fp544037857@gmail.com>

https://pkg.go.dev/testing#T.Fatal 
> Fatal is equivalent to Log followed by FailNow.

so `break` will not run.


or-else:
Thanks!

PR #783, opened 2022-08-23
More correct handling of channel topics in cluster mode

aforge:
Errors fixed:
1. When both `chnX` and `grpX` multiplexing sessions (for the same channel topic) are attached to topic master, send messages to only `grpX`.
2. When rewriting topic name for the client (`Topic.maybeFixTopicName()`) take subscription type (channel/non-channel) into account.
3. Set `Original` field in the `leave` messages from topic proxy to master so the master can process these requests correctly.

### DIFF
--- a/server/topic.go
+++ b/server/topic.go
@@ -241,18 +241,14 @@ func (t *Topic) userIsReader(uid types.Uid) bool {
 	return (modeGiven & modeWant).IsReader()
 }
 
-// maybeFixTopicName sets the topic field in `msg` depending on the uid and subscription type.
-func (t *Topic) maybeFixTopicName(msg *ServerComMessage, uid types.Uid, isChanSub bool) {
-	// For zero uids we don't know the proper topic name.
-	if uid.IsZero() {
-		return
-	}
+// prepareBroadcastableMessage sets the topic field in `msg` depending on the uid and subscription type.
+func (t *Topic) prepareBroadcastableMessage(msg *ServerComMessage, uid types.Uid, isChanSub bool) {
 	// We are only interested in broadcastable messages.
 	if msg.Data == nil && msg.Pres == nil && msg.Info == nil {
 		return
 	}
 
-	if t.cat == types.TopicCatP2P || (t.cat == types.TopicCatGrp && t.isChan) {
+	if (t.cat == types.TopicCatP2P && !uid.IsZero()) || (t.cat == types.TopicCatGrp && t.isChan) {
 		// For p2p topics topic name is dependent on receiver.
 		// Channel topics may be presented as grpXXX or chnXXX.
 		var topicName string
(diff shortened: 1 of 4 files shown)


server/topic.go
```diff
@@ -241,23 +241,33 @@ func (t *Topic) userIsReader(uid types.Uid) bool {
 	return (modeGiven & modeWant).IsReader()
 }
 
-// maybeFixTopicName sets the topic field in `msg` depending on the uid.
-func (t *Topic) maybeFixTopicName(msg *ServerComMessage, uid types.Uid) {
+// maybeFixTopicName sets the topic field in `msg` depending on the uid and subscription type.
+func (t *Topic) maybeFixTopicName(msg *ServerComMessage, uid types.Uid, isChanSub bool) {
 	// For zero uids we don't know the proper topic name.
 	if uid.IsZero() {
```
or-else, line 247: Not sure is `uid` could be zero for channels, but this check is only valid for p2p topics. So, maybe make the first check
`if (t.cat == types.TopicCatP2P && !uid.IsZero()) || (t.cat == types.TopicCatGrp && t.isChan) {`

server/topic.go
```diff
@@ -1269,18 +1279,32 @@ func (t *Topic) broadcastToSessions(msg *ServerComMessage) {
...
 
 		// Topic name may be different depending on the user to which the `sess` belongs.
-		t.maybeFixTopicName(msg, pssd.uid)
+		t.maybeFixTopicName(msg, pssd.uid, pssd.isChanSub)
 
+		msgCopy := msg.copy()
 		// Send channel messages anonymously.
-		if pssd.isChanSub && msg.Data != nil {
-			msg.Data.From = ""
+		if pssd.isChanSub && msgCopy.Data != nil {
```
or-else, line 1302: Consider moving this logic to `maybeFixTopicName` and renaming it to `prepareMultiplexMessage` or some such.

server/topic_proxy.go
```diff
@@ -129,14 +129,29 @@ func (t *Topic) handleProxyLeaveRequest(msg *ClientComMessage, killTimer *time.T
...
 		// of its multiple hosted sessions to delete.
 		msg.AsUser = asUid.UserId()
 		msg.Leave = &MsgClientLeave{}
 		msg.init = true
 	}
+	// Make sure we set the Original field if it's empty (e.g. when session is terminating altogether).
+	if msg.Original == "" {
+		if t.cat == types.TopicCatGrp && t.isChan {
+			// It's a channel topic. Original topic name depends the subscription type.
+			var toriginal string
```
or-else, line 144: This looks redundant. You can direct assign to `msg.Original` instead of assigning to `toriginal` first.

or-else (approved):

server/topic.go

server/topic.go
aforge: Renamed `prepareBroadcastableMessage()`.

server/topic_proxy.go
aforge: Done.

Issue #786, opened 2022-08-31
Server went in deadlock in cluster setup


aman-gupta-doc:
### Deadlock Condition Observed
There is a cluster of 3 nodes. During high load on these servers with tinode running, the servers went in a deadlock condition because  the goroutine in context of `hub.go` got stuck in RPC call to another server within the cluster. This condition is noticeable for all the nodes in the cluster, hence creating a circular dependencies of goroutines leading to deadlock.

#### Server-side
- [ ] web.tinode.co, api.tinode.co
- [ ] sandbox.tinode.co
- [x] Your own setup:
  * Platform: Linux
  * Version: 0.19.3
  * DB: MongoDB
  * cluster: `true` (3 nodes)

### Steps to reproduce
In the Scenaro, there are 3 Servers, S1, S2, S3 respectively. The servers when put in high load the channel `globals.hub.routeSrv` got full and went in block state (_buffered channel blocking when the buffer is full_) and at this time if hub of S1 makes a call to the function `globals.cluster.routeToTopicIntraCluster`  then this function makes a request to  `Cluster.Route` of S2. This results in S1's hub waiting for response but S2's channel `globals.hub.routeSrv` is in a block state (_as mentioned above_). Also, S1's hub is stuck and this scenario happens with all three nodes at the same time and hence all servers went in a deadlock situation which it does not  recover from until it is restarted.

https://github.com/tinode/chat/blob/ebcb536d79b57fe04cfde5b239dd768c798a9c17/server/cluster.go#L584

For verifying this, I have attached goroutine stacks when cluster went in deadlock

**Hub GoRoutine stack**:
S1
```
goroutine 51 [chan receive, 38 minutes]:
net/rpc.(*Client).Call(...)
	/usr/local/go/src/net/rpc/client.go:321
main.(*ClusterNode).call(0xc000160960, {0x111d390, 0xd}, {0xf018c0, 0xc00d7edb40}, {0xeeee80, 0xc00da4604d})
	/Users/gene/go/src/github.com/tinode/chat/server/cluster.go:286 +0xfe
main.(*ClusterNode).route(...)
	/Users/gene/go/src/github.com/tinode/chat/server/cluster.go:381
main.(*Cluster).routeToTopicIntraCluster(0xc000092410, {0xc00ca5e650, 0xe}, 0xc00ca6ab40, 0x0)
	/Users/gene/go/src/github.com/tinode/chat/server/cluster.go:843 +0x1a5
[...]

PR #787, opened 2022-09-05
Remove deadlock on server overload, #786

or-else:
1. Added a network connection timeout when connecting to nodes.
2. Added a channel+goroutine and a check for overload for RPC calls from Proxy to Master. Proxy topic sends a message to clients if RCP call to master failed. 
3. Added a check for overload when master node sends requests to master's hub.

What I don't like here is that when the master rejects a call from the proxy, the error is just logged by the `p2mSenderLoop`, while it should be reported back to the calling session. It's highly unlikely that this would happen in production because the buffer cluster->hub at master is much bigger than the buffer at proxy->master. But still.

### DIFF
--- a/server/cluster.go
+++ b/server/cluster.go
@@ -20,10 +20,18 @@ import (
 )
 
 const (
+	// Network connection timeout.
+	clusterNetworkTimeout = 3 * time.Second
 	// Default timeout before attempting to reconnect to a node.
-	defaultClusterReconnect = 200 * time.Millisecond
+	clusterDefaultReconnectTime = 200 * time.Millisecond
 	// Number of replicas in ringhash.
 	clusterHashReplicas = 20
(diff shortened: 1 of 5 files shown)


server/cluster.go
```diff
@@ -366,6 +391,16 @@ func (n *ClusterNode) proxyToMaster(msg *ClusterReq) error {
 	return err
 }
 
+// proxyToMaster forwards request from topic proxy to topic master.
+func (n *ClusterNode) proxyToMasterAsync(msg *ClusterReq) error {
```
aforge, line 395: I don't see any calls to this method.
Can you double check please?

or-else: Oops. Fixed.

aforge (approved):

PR #796, opened 2022-11-09
Add document for own API key

SMAH1:
After install my tinode, I seach/read/google forum and github to define correct `API Key`. I decided to save my experience.

### DIFF
--- a/docs/API_KEY.md
+++ b/docs/API_KEY.md
@@ -2,7 +2,7 @@
 
 ## What is API key?
 
-Trinode encrypts data for security. For this purpose, two keys are used:
+These keys are used to generate an API key which provides some minor protection from automated tools scraping the API. For this purpose, two keys are used:
 
 * `HMAC salt` that is used on the server side to verify the API key.
 * `API key` that is used on the client side.
@@ -56,4 +56,6 @@ const API_KEY = 'AQAAAAAFAAAqXLPhti0JwhyVXBWKXTPW';
 ....
 
 ```
+
+*Note*: The preferred way is to edit the source file then recompile.
 
--- a/docs/API_KEY.md
+++ b/docs/API_KEY.md
@@ -7,11 +7,50 @@ Trinode encrypts data for security. For this purpose, two keys are used:
 * `HMAC salt` in server side. Hereafter referred to as _Server Key_.
 * `API key` in client side. Hereafter referred to as _Client Key_.
 
-## Prerequisite
+## Generate Server/Client Keys
+
+### 1) Generate Automatically
+
+Use the [keygen](https://github.com/tinode/chat/blob/master/keygen/README.md) which is available in the tinode folder. Run this program with the `-salt auto` option. As below:
+
+```bash
+./keygen -salt auto
+```
+
+result like:
+
+```text
+API key v1 seq1 [ordinary]: AQAAAAABAACGOIyP2vh5avSff5oVvMpk
+Used HMAC salt: TC0Jzr8f28kAspXrb4UYccJUJ63b7CSA16n1qMxxGpw=
+```
+
+There are two lines:
+
+* In first line detect _Client Key_. In this example is `AQAAAAABAACGOIyP2vh5avSff5oVvMpk`.
+* In second line detect _Server Key_. In this example is `TC0Jzr8f28kAspXrb4UYccJUJ63b7CSA16n1qMxxGpw=`.
+
+*Note*: This value is random and will be different for you.
+*Note*: You can use the `-sequence` option to make the key more complex. like:
+
+```bash
+./keygen -salt TC0Jzr8f28kAspXrb4UYccJUJ63b7CSA16n1qMxxGpw= -sequence 5
+```
+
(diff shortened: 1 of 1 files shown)


docs/API_KEY.md
```diff
@@ -0,0 +1,112 @@
+# How to add new/own API key to tinode?
+
+## What is API key?
+
+Trinode encrypts data for security. For this purpose, two keys are used:
+
+* `HMAC salt` in server side. Hereafter referred to as _Server Key_.
+* `API key` in client side. Hereafter referred to as _Client Key_.
+
+## Prerequisite
```
or-else, line 10: This section is not needed. Run `keygen --salt=auto` and it will auto-generate cryptographically-secure random salt.

docs/API_KEY.md
```diff
@@ -0,0 +1,112 @@
...
+```
+
+Note: This value is random and will be different for you.
+
+## Build Server/Client Keys from your key
+
+Now you need to use the key created in the previous step to create the two keys needed by tinode. Use the [keygen](https://github.com/tinode/chat/blob/master/keygen/README.md) which is available in the tinode folder. Run this program with the `-salt` option and give it the key. As below:
+
+```bash
+./keygen -salt TC0Jzr8f28kAspXrb4UYccJUJ63b7CSA16n1qMxxGpw=
```
or-else, line 57: Just `./keygen -salt auto`

docs/API_KEY.md
```diff
@@ -0,0 +1,112 @@
...
+
+```
+
+Then restart tinode.
+
+## Add Client Key to Tinode
+
+### Add to WebApp
+
+Open file `index.*.js` in the _tinode/static/umd_ folder. Find `API_KEY` variable and change it. Like:
```
or-else, line 104: The preferred way is to edit the source file then recompile. It should be mentioned.

docs/API_KEY.md
```diff
@@ -0,0 +1,151 @@
...
+Used HMAC salt: TC0Jzr8f28kAspXrb4UYccJUJ63b7CSA16n1qMxxGpw=
+```
+
+There are two lines:
+
+* In first line detect _Client Key_. In this example is `AQAAAAABAACGOIyP2vh5avSff5oVvMpk`.
+* In second line detect _Server Key_. In this example is `TC0Jzr8f28kAspXrb4UYccJUJ63b7CSA16n1qMxxGpw=`.
+
+*Note*: This value is random and will be different for you.
+*Note*: You can use the `-sequence` option to make the key more complex. like:
```
or-else, line 33: Sequence is not used for making the key more complex. It's intended for key revocation but currently unused. Please remove this paragraph.

docs/API_KEY.md
```diff
@@ -0,0 +1,151 @@
...
+
+```text
+API key v1 seq1 [ordinary]: AQAAAAABAACGOIyP2vh5avSff5oVvMpk
+Used HMAC salt: TC0Jzr8f28kAspXrb4UYccJUJ63b7CSA16n1qMxxGpw=
+```
+
+There are two lines:
+
+* In first line detect _Client Key_. In this example is `AQAAAAABAACGOIyP2vh5avSff5oVvMpk`.
+* In second line detect _Server Key_. In this example is `TC0Jzr8f28kAspXrb4UYccJUJ63b7CSA16n1qMxxGpw=`.
```
or-else, line 30: Let's call HMAC salt `HMAC salt` rather than _Server key_. You can mention that it's used on the server side to verify the API key.

docs/API_KEY.md
```diff
@@ -0,0 +1,151 @@
...
+
+result is:
+
+```text
+API key v1 seq5 [ordinary]: AQAAAAAFAAAqXLPhti0JwhyVXBWKXTPW
+Used HMAC salt: TC0Jzr8f28kAspXrb4UYccJUJ63b7CSA16n1qMxxGpw=
+```
+
+
+### 2) Generate Manually
```
or-else, line 47: I'm not sure this section is needed at all. I will probably remove the ability to use external salt as it brings no value. It's there for historical reasons only.

docs/API_KEY.md
or-else: Up

or-else: Please mention adding key to the source file.

or-else: This has not been resolved.

docs/API_KEY.md
```diff
@@ -0,0 +1,59 @@
+# How to add new/own API key to tinode?
+
+## What is API key?
+
+Trinode encrypts data for security. For this purpose, two keys are used:
```
or-else, line 5: Actually, these keys are not used for encryption. These keys are used to generate an API key which provides some minor protection from automated tools scraping the API.

or-else: Also `Trinode`?

or-else:
I copied some of your comments to https://github.com/tinode/chat/tree/next/keygen
I don't think this PR is needed now. Thank you!

Issue #795, opened 2022-11-09
In init-db, Ignore tags of user


SMAH1:
I install last version of tinode (version 0.20) in my server:
OS: Debian 10
DB: Mysql 8

I build `mydata.json` like this:

```json
{
    "users": [
        {
			"createdAt": "-140h",
			"email": "smah1@example.com",
			"passhash": "111111",
			"trusted": {"verified": true},
			"tags": ["myuser"],
			"state": "ok",
			"status": "",
			"username": "smah1",
			"addressBook": []
		}
    ]
}
```

I define one user with tag `myuser`.

I run command `./init-db -data=mydata.json`. Result is OK, like :

```bash
2022/11/09 10:31:49 Database adapter: 'mysql'; version: 112
2022/11/09 10:31:49 Database not found. Creating.
2022/11/09 10:31:51 Database initialized
2022/11/09 10:31:51 Generating users...
usr;smah1;usr9N4rfMR1bJA;111111
2022/11/09 10:31:51 Generating group topics...
2022/11/09 10:31:51 Generating P2P subscriptions...
2022/11/09 10:31:51 Generating group subscriptions...
2022/11/09 10:31:51 All done.
```

Now start tinode. It's OK.

Now when see information of my user in webapp and db (by phpmyadmin), there is not tag `myusr`. but add two tags `basic:smah1` and `email:smah1@example.com`.

It seems that the program ignores the input tags

![image](https://user-images.githubusercontent.com/21136680/200877904-7a26c4b8-b729-4f59-a1d7-67d8c223f154.png)

or-else:
Thanks for the bug report and the fix. Closing.

PR #797, opened 2022-11-09
Fix Bug: When Init-DB by json file, ignore tags of user if defined (Issue #795)

SMAH1:
When use `./init-db -data=mydata.json`, It seems that the program ignores the input tags. #795 

I dont programmer of *GO* but I think that is fixed. *Please review it*


or-else:
Thanks, that's the right fix.

PR #802, opened 2022-11-10
Refactoring: Move some structures used in tinode.conf

SMAH1:
I think project need some refactoring.

### DIFF
--- a/server/config_types.go
+++ b/server/config_types.go
@@ -1,0 +1,38 @@
+
+/******************************************************************************
+ *
+ *  Description :
+ *
+ *    All structure defined in tinode.conf
+ *
+ *****************************************************************************/
+
+ package main
+
+ import (
+	"encoding/json"
+)
+
+// Credential validator config.
+type validatorConfig struct {
+	// TRUE or FALSE to set
+	AddToTags bool `json:"add_to_tags"`
+	//  Authentication level which triggers this validator: "auth", "anon"... or ""
+	Required []string `json:"required"`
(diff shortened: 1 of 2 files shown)


or-else:
I'm not sure this is needed. Please explain.

rkgarcia:
Sounds as a better file organization approach

SMAH1:
1) There are some bug (or mistake) in tinode-db. For example : `basic:USERNAME` to tag when use `-data=...`, But the value of `validatorConfig :add_to_tag` is ignored.
validatorConfig  define in `main`. Do that import? or Do Recode tinode-db?
2) I think reorganize code is better way.

or-else:
1. I don't understand. Please explain or show an example.
2. It's basically a matter of opinion/personal preference. Let's leave it as is.

or-else:
1. I think I understand now. `tinode-db` uses a subset of config values compare to the main server:
https://github.com/tinode/chat/blob/master/tinode-db/tinode.conf
`tinode-db` does not use `validatorConfig`, etc. The `add_to_tag` is hardcoded to `true`:
https://github.com/tinode/chat/blob/1aa7942a29aff5cade9193be569deabcc49a8860/tinode-db/gendb.go#L30

The purpose of `tinode-db` is to initialize the database and maybe load sample data. It's not intended for managing users in production.

SMAH1:
#805

or-else:
OK, let's continue the conversation in https://github.com/tinode/chat/issues/805. Closing.

PR #811, opened 2022-12-03
fix: wrong comment

YiuTerran:
should use ";" instead of ":" in comment


or-else:
Good catch. Thank you!

PR #813, opened 2022-12-06
Stale unvalidated account garbage collection.

aforge:
Attempts to identify stale user accounts with
required unvalidated credentials and clean them up.
Note: only MySQL functionality implemented.

### DIFF
--- a/server/user.go
+++ b/server/user.go
@@ -2,6 +2,8 @@ package main
 
 import (
 	"container/heap"
+	"math/rand"
+	"strings"
 	"time"
 
 	"github.com/tinode/chat/server/auth"
@@ -149,7 +151,7 @@ func replyCreateUser(s *Session, msg *ClientComMessage, rec *auth.Rec) {
 			globals.authValidators[rec.AuthLevel], s.sid)
 		// Attempt to delete incomplete user record
 		store.Users.Delete(user.Uid(), false)
-		_, missing := stringSliceDelta(globals.authValidators[rec.AuthLevel], credentialMethods(creds))
+		_, missing, _ := stringSliceDelta(globals.authValidators[rec.AuthLevel], credentialMethods(creds))
 		s.queueOut(decodeStoreError(types.ErrPolicy, msg.Id, "", msg.Timestamp,
 			map[string]interface{}{"creds": missing}))
 		return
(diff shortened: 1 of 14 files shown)


server/user.go
```diff
@@ -1143,3 +1146,55 @@ func userUpdater() {
...
+						uid := uids[i]
+						authLvl := authLvls[i]
+						cred := strings.Split(creds[i], ",")
+						_, _, missing := stringSliceDelta(globals.authValidators[authLvl], cred)
+						if len(missing) > 0 {
+							uidsToDelete = append(uidsToDelete, uid)
+						}
+					}
+					logs.Info.Println("Stale account GC will delete the following uids", uidsToDelete)
+					if len(uidsToDelete) > 0 && !dryRun {
```
or-else, line 1183: I think the `logs.Info.Println("Stale account GC will delete the following uids")` from line 1182 should be moved under this if, otherwise it will add a useless constant log entry most of the time.

BTW, I'm not sure `dryRun` is needed for anything but debugging. Maybe remove it altogether?

server/user.go
```diff
@@ -1143,3 +1146,55 @@ func userUpdater() {
...
+						for _, uid := range uidsToDelete {
+							if err := store.Users.Delete(uid, true); err != nil {
+								logs.Warn.Printf("Could not delete user %s: %w", uid.UserId(), err)
+							}
+						}
+					}
+				} else {
+					logs.Warn.Println("User account gc error: %+v", err)
```
or-else, line 1191: Either `Printf("User account gc error: %+v", err)`, or `Println("User account gc error:", err)`

server/utils.go
```diff
@@ -106,15 +106,16 @@ func normalizeTags(src []string) types.StringSlice {
 // stringDelta extracts the slices of added and removed strings from two slices:
 //   added :=  newSlice - (oldSlice & newSlice) -- present in new but missing in old
 //   removed := oldSlice - (oldSlice & newSlice) -- present in old but missing in new
-func stringSliceDelta(rold, rnew []string) (added, removed []string) {
+//   intersection := oldSlice & newSlice -- present in both old and new
+func stringSliceDelta(rold, rnew []string) (added, removed, intersection []string) {
```
or-else, line 110: It would be nice to add a test.

or-else (approved):

or-else:
Let's merge it to next instead of devel. I think it needs to be tested first because it can do damage.

server/utils.go
aforge: Done.

aforge (commented):
Working on remaining adapters now.

PR #823, opened 2023-01-18
Update faq.md

theslash:
Changed ./tinode-db to ./init-db for creating a root user


or-else:
Unfortunately it's not that simple. When you build from sources then the name of the binary is `tinode-db`. This change will just make confusion happen in a different place.

theslash:
ok. thanks for the info. perhaps it should be noted that at least at the moment the binary in the docker containers is called init-db. or will that change in the future?

PR #827, opened 2023-01-31
Fix deletion of unvalidated accounts

or-else:
The databases would have to be updated too: the query requires new indexes. I'll update the DBs together with the change to `kvmeta`.

### DIFF
--- a/server/user.go
+++ b/server/user.go
@@ -1156,25 +1156,26 @@ func garbageCollectUsers(period time.Duration, blockSize, minAccountAgeHours int
 	go func() {
 		// Add some randomness to the tick period to desynchronize runs on cluster nodes:
 		// 0.75 * period + rand(0, 0.5) * period.
-		period = (period >> 1) + (period >> 2) + time.Duration(rand.Intn(int(period>>1)))
+		period = period - (period >> 2) + time.Duration(rand.Intn(int(period>>1)))
 		gcTicker := time.Tick(period)
-		logs.Info.Printf("Starting account GC with period %s, block size %d and minimum account age %d hours",
-			period, blockSize, minAccountAgeHours)
+		logs.Info.Printf("Stale account GC started with period %s, block size %d, min account age %d hours",
+			period.Round(time.Second), blockSize, minAccountAgeHours)
 		staleAge := time.Hour * time.Duration(minAccountAgeHours)
 		for {
 			select {
 			case <-gcTicker:
+				logs.Info.Println("Stale account GC ticker")
 				if uids, err := store.Users.GetUnvalidated(time.Now().Add(-staleAge), blockSize); err == nil {
 					if len(uids) > 0 {
-						logs.Info.Println("Stale account GC will delete the following uids:", uids)
+						logs.Info.Println("Stale account GC will delete uids:", uids)
 						for _, uid := range uids {
 							if err = store.Users.Delete(uid, true); err != nil {
-								logs.Warn.Printf("Could not delete user %s: %+v", uid.UserId(), err)
+								logs.Warn.Printf("Stale account GC failed to delete %s: %+v", uid.UserId(), err)
 							}
 						}
 					}
 				} else {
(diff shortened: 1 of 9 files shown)


server/user.go
```diff
@@ -1157,36 +1156,25 @@ func garbageCollectUsers(period time.Duration, blockSize, minAccountAgeHours int
 	go func() {
 		// Add some randomness to the tick period to desynchronize runs on cluster nodes:
 		// 0.75 * period + rand(0, 0.5) * period.
-		period = (period >> 1) + (period >> 2) + time.Duration(rand.Intn(int(period>>1)))
+		period = period - (period >> 2) + time.Duration(rand.Intn(int(period>>1)))
```
aforge, line 1159: This is `0.5 * period + rand(0, 0.5) * period`. It's guaranteed to be less than `period`.
Mean = 0.75 * period, stddev = 0.14 * period.
Ideally we should make mean = period which the previous distribution did.

or-else: The result is a uniformly distributed value [0.75 * period, 1.25 * period]:
```
period = period - (period >> 2) + time.Duration(rand.Intn(int(period>>1)))
```
==
```
1 - 0.25 + rand(0, 0.5)
```

The `rand.Intn(int(period>>1))` is _uniformly_ not _normally_ distributed. It has a math expectation of 0.25 * period.

or-else: `(period >> 1) + (period >> 2) === period - (period >> 2)`
`0.5 + 0.25 === 1 - 0.25`

aforge: My bad. You are right. The shift confused me.

aforge:
Got this from mysql:
```
W2023/01/31 17:37:22 user.go:1177: Stale account GC error: Error 1064: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near '?GROUP BY u.id, u.updatedat HAVING total=0 ORDER BY u.updatedat ASC LIMIT ?' at line 1
```

or-else:
> Got this from mysql:
> 
> ```
> W2023/01/31 17:37:22 user.go:1177: Stale account GC error: Error 1064: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near '?GROUP BY u.id, u.updatedat HAVING total=0 ORDER BY u.updatedat ASC LIMIT ?' at line 1
> ```

Fixed. I reformatted the line and did not check after that.

aforge:
RethinkDB & MongoDB are fine.

Unrelated to this PR. Uploading a large (out-of-band) avatar in RethinkDB during signup gets this:
```
W2023/01/31 17:42:40 user.go:176: create user: failed to link avatar attachment rethinkdb: No attribute `Attachments` in object:
{
	"Access":	{
		"Anon":	0,
		"Auth":	63
	},
	"CreatedAt":	{
		"$reql_type$":	"TIME",
		"epoch_time":	1675215760.231,
		"timezone":	"+00:00"
	},
	"Devices":	null,
	"Id":	"IDY1N0g0jOc",
	"LastSeen":	null,
	"Public":	{
		"fn":	"test",
		"photo":	{
			"data":	"␡",
			"ref":	"/v0/file/s/shkSlOErX4Q.png",
			"type":	"jpeg"
		}
	},
	"State":	0,
	"Tags":	[
		"basic:test"
	],
	"Trusted":	null,
	"UpdatedAt":	{
		"$reql_type$":	"TIME",
		"epoch_time":	1675215760.231,
		"timezone":	"+00:00"
	},
	"UserAgent":	""
} in:
r.DB("tinode").Table("users").Get("IDY1N0g0jOc").Field("Attachments") p32ka4kVxbk
```

aforge (approved):

or-else:
> Unrelated to this PR. Uploading a large (out-of-band) avatar in RethinkDB during signup gets this:

Fixed in https://github.com/tinode/chat/commit/e57d843c2ee9e3d158bc2c0708da2e60ebededde

PR #828, opened 2023-02-01
Support for password reset over SMS for Rethink + Mongo

or-else:
Also, DB versions bumped because of indexes needed for stale account GC.


server/validate/tel/validate.go
```diff
@@ -2,58 +2,207 @@
...
+	code, err := rand.Int(rand.Reader, v.maxCodeValue)
+	if err != nil {
+		return false, err
+	}
+	resp = strconv.FormatInt(code.Int64(), 10)
+	resp = strings.Repeat("0", v.CodeLength-len(resp)) + resp
+
+	var template *textt.Template
+	if v.langMatcher != nil {
+		_, idx := i18n.MatchStrings(v.langMatcher, lang)
```
aforge, line 157: What if `lang` doesn't match?
This may lead to a crash. We need a fallback mechanism to the default template.

server/validate/email/validate.go
```diff
@@ -365,12 +271,12 @@ func (v *validator) Request(user t.Uid, email, lang, resp string, tmpToken []byt
...
+	code, err := crand.Int(crand.Reader, v.maxCodeValue)
 	if err != nil {
 		return false, err
 	}
 	resp = strconv.FormatInt(code.Int64(), 10)
-	resp = strings.Repeat("0", codeLength-len(resp)) + resp
+	resp = strings.Repeat("0", v.CodeLength-len(resp)) + resp
 
 	var template *textt.Template
 	if v.langMatcher != nil {
```
aforge, line 282: Need fallback mechanism in case `lang` doesn't match any available templates.

aforge:
Rethinkdb won't build b/c of syntax errors. Here's a fix (but please double check):

diff --git a/server/db/rethinkdb/adapter.go b/server/db/rethinkdb/adapter.go
index 3d49463d..4ad5c3ae 100644
--- a/server/db/rethinkdb/adapter.go
+++ b/server/db/rethinkdb/adapter.go
@@ -2799,7 +2799,7 @@ func (a *adapter) PCacheUpsert(key string, value string, failOnDuplicate bool) e
                action = "update"
        }
 
-       _, err := rdb.DB(a.dbName).Table("kvmeta").Insert(doc, InsertOpts{Conflict: action}).RunWrite(session)
+       _, err := rdb.DB(a.dbName).Table("kvmeta").Insert(doc, rdb.InsertOpts{Conflict: action}).RunWrite(a.conn)
        if rdb.IsConflictErr(err) {
                return t.ErrDuplicate
        }
@@ -2819,8 +2819,8 @@ func (a *adapter) PCacheExpire(keyPrefix string, olderThan time.Time) error {
                return t.ErrMalformed
        }
 
-       _, err = rdb.DB(a.dbName).Table("kvmeta").
-               Filter(rdb.Row.Field("CreatedAt").Lt(olderThan).And(rdb.Row.Field("key").Match("^"+keyPrefix)))).
+       _, err := rdb.DB(a.dbName).Table("kvmeta").
+               Filter(rdb.Row.Field("CreatedAt").Lt(olderThan).And(rdb.Row.Field("key").Match("^"+keyPrefix))).
                Delete().
                RunWrite(a.conn)

server/validate/tel/validate.go
or-else: Then the default language is used. That's the first language in the list of languages passed to the `NewMatcher`. That's `en` in our config and the template for English is provided.

server/validate/email/validate.go
or-else: It's automatic. The matcher will use the first language from the list when all else fails.

server/tinode.conf
```diff
@@ -385,14 +385,36 @@
...
 			"add_to_tags": true,
 			"config": {
-				"languages": ["en"],
-				"template": "./templ/sms-validation.templ",
-				"max_retries": 4,
+				// Address of the host where the Tinode server is running. This will be used
+				// in URLs in the SMS.
+				"host_url": "http://localhost:6060/",
+
+				// Optional list of human languages to try to load templates for. If you don't care about i18n,
```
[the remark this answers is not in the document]
or-else: done

or-else:
> Rethinkdb won't build b/c of syntax errors. Here's a fix (but please double check):

Fixed

aforge (approved):

PR #831, opened 2023-02-19
fix: add quotes to str in tn-cli : encode_to_bytes

queses:
Fixes "resolve" and "pub" commands in CLI.

Examples of wrong behaviour before fix:

**pub**:
image

**resolve**:
image


or-else:
Forgot to mention and made a mistake myself: please send PRs against `devel`, not `master`.

PR #834, opened 2023-02-25
fix: add message id where possible in session.go:dispatch

queses:


or-else:
Sorry, I was wrong. This PR does nothing: the `msg.Id` is not yet assigned and consequently you are still assigning `""`.

or-else:
The assignments happen later, for instance here:

https://github.com/tinode/chat/pull/834/files#diff-7cab63ea28de45f2ae502528a434695d39c1d55189b56919083dba41f355b025L539

PR #844, opened 2023-03-20
Add "sender" header for messages sent on_behalf_of

or-else:
* Removed `hashtags` and `priority` from the README list of possible message headers because they are not going to be implemented ever.
* When replacing webRTC messages, use existing message `head` as basis when available instead of always creating a new `head`.
* Implemented `sender` header for messages sent `on_behalf_of`.

### DIFF
--- a/server/calls.go
+++ b/server/calls.go
@@ -171,6 +171,8 @@ func initVideoCalls(jsconfig json.RawMessage) error {
 	return nil
 }
 
+// Add webRTC-related headers to message Head. The original Head may already contain some entries,
+// like 'sender', preserve them.
 func (call *videoCall) messageHead(head map[string]interface{}, newState string, duration int) map[string]interface{} {
 	if head == nil {
 		head = map[string]interface{}{}


server/calls.go
```diff
@@ -171,11 +171,14 @@ func initVideoCalls(jsconfig json.RawMessage) error {
 	return nil
 }
 
-func (call *videoCall) messageHead(newState string, duration int) map[string]interface{} {
-	head := map[string]interface{}{
-		"replace": ":" + strconv.Itoa(call.seq),
-		"webrtc":  newState,
+func (call *videoCall) messageHead(head map[string]interface{}, newState string, duration int) map[string]interface{} {
+	if head == nil {
+		head = map[string]interface{}{}
```
aforge, line 176: Any particular need for this?

server/topic.go
```diff
@@ -964,6 +964,17 @@ func (t *Topic) saveAndBroadcastMessage(msg *ClientComMessage, asUid types.Uid,
 		}
 	}
 
+	if msg.sess != nil && msg.sess.uid != asUid {
+		// Add 'sender' header indicating that message was sent on behalf of another user.
```
aforge, line 968: Can you change the comment to state that:
`sender` is the user who actually sent the message on behalf of `asUid`.

docs/API.md
```diff
@@ -881,10 +881,8 @@ The following values are currently defined for the `head` field:
  * `attachments`: an array of paths indicating media attached to this message `["/v0/file/s/sJOD_tZDPz0.jpg"]`.
  * `auto`: `true` when the message was sent automatically, i.e. by a chatbot or an auto-responder.
  * `forwarded`: an indicator that the message is a forwarded message, a unique ID of the original message, `"grp1XUtEhjv6HND:123"`.
- * `hashtags`: an array of hashtags in the message without the leading `#` symbol: `["onehash", "twohash"]`.
  * `mentions`: an array of user IDs mentioned (`@alice`) in the message: `["usr1XUtEhjv6HND", "usr2il9suCbuko"]`.
  * `mime`: MIME-type of the message content, `"text/x-drafty"`; a `null` or a missing value is interpreted as `"text/plain"`.
- * `priority`: message display priority: hint for the client that the message should be displayed more prominently for a set period of time; only `"high"` is currently defined; `{"level": "high", "expires": "2019-10-06T18:07:30.038Z"}`; `priority` can be set by the topic owner or administrator (`A` permission) only. The `"expires"` qualifier is optional.
```
aforge, line 887: Out of curiosity, what was the use case for this one?

aforge (approved):

server/calls.go
or-else: If a call was initiated by a root user on behalf of an ordinary user, the `sender` header should be preserved.

docs/API.md
or-else: That was the original idea how to make pinned/high priority messages.

server/calls.go
aforge: I see. I think it's worth leaving a comment here.

aforge (approved):

server/topic.go
or-else: Done

server/calls.go
or-else: Comment added.

PR #855, opened 2023-04-06
Postgres adapter

or-else:
Cleanup and a couple of bug fixes.


server/db/postgres/adapter.go
```diff
@@ -0,0 +1,3543 @@
+//go:build postgres
+// +build postgres
+
+// Package postgres is a database adapter for PostgreSQL.
+package postgres
+
+import (
+	//"bytes"
```
aforge, line 8: Remove.

aforge (commented):
Do you think you can add docker compose configs in [docker](https://github.com/tinode/chat/tree/master/docker)/docker-compose/?

or-else:
> Do you think you can add docker compose configs in [docker](https://github.com/tinode/chat/tree/master/docker)/docker-compose/?

Added the files, but I cannot test them:

1. The files are for amd64, I'm on arm.
2. The `tinode-postgres` container has to be built and published first.

server/db/postgres/adapter.go
or-else: Done

aforge (approved):

PR #863, opened 2023-05-20
add a pyi file generation for tinode_grpc module

Ovizro:
[Stub files(.pyi)](https://peps.python.org/pep-0484/#stub-files) can make users get auto-completion prompts when writing programs.


or-else:
Thanks!

Issue #866, opened 2023-05-27
Topic stuck for receiving user in p2p/grp


dilshansdoq:
### Subject of the issue
Hi,
This seems like a bug. So we have slightly changed the client implementation such that if a user receives `pres` with `what:msg` indicating someone has messaged this user, will fetch that message and mark delivery (sub:topic:get:what:data -> note:recv -> leave:topic), and this will happen for each message pres received.
After some time of messaging, topic gets stuck for the user at receiving end of messages, such that user is not receiving any pres message(indicating not subbed to topic) nor any data messages(indicating subbed to topic). This user is not receiving anything happening in this topic. Also <server>/debug/status show user is already subscribed to topic. 
This gets fixed if user sends another leave:topic request. 

#### Server-side
- [x] Your own setup:
  * platform: Linux
  * version of Tinode server: 0.21 + (some cherry picks from devel)
  * database backend: mongodb
  * standalone

#### Client-side
- [x] TinodeWeb/tinodejs: javascript client
  * Browser make and version: Microsoft edge Version 113.0.1774.35 (Official build) (64-bit)
- [x] Tindroid: Android app
  * Android API level 30,31,32,33
  * Hardware: Latest 2022-2023 flagships
- Version of the client, e.g. `0.21.0`
- [x] Your own client. Describe it:
  * Transport: ws
  * Testing script with NodeJS 

### Steps to reproduce
This issue isn't 100% reproducible. It takes many tries and happens randomly (my intuition is something is happening when subbing and leaving). However i have written a script replicating what would happen when a user (on receiving end) receives pres:what:msg and does (sub:topic:get:what:data -> note:what:recv -> leave:topic). 
[topic-stuck-issue.zip](https://github.com/tinode/chat/files/11582990/topic-stuck-issue.zip)

### Expected behaviour
User should be able to receive either {pres} or {data}.

### Actual behaviour
User is not receiving neither {pres} nor {data}

### Server-side log
[topic-stuck-server.log](https://github.com/tinode/chat/files/11582970/topic-stuck-server.log)

### Client-side log
Receving side: 
I have manually put some comments where topic was stuck for receiving user and was not receiving any updates.
[topic-stuck-client.log](https://github.com/tinode/chat/files/11582972/topic-stuck-client.log)

or-else:
@aforge This sounds like something related to `Session.inflightReqs`. Would you take a look?

dilshansdoq:
Hi, any update on this issue?

[...]

PR #878, opened 2023-07-07
Fix a race condition around Session.subs.

aforge:
1. Session.addSub & Session.delSub should be happening (atomically) together modifications of Topic.sessions.
2. Only one of Session.subscribe() and Session.leave() is allowed to be inflight at any given moment.

Attempt at addressing https://github.com/tinode/chat/issues/866


or-else (approved):

or-else:
> It looks like the `wg sync.WaitGroup` is now redundant. The `sem` portion duplicates its functionality. Or am I missing something?

Actually no, `wg` is still needed.

PR #882, opened 2023-08-16
fixed missing pyi files in python wheel

Ovizro:


or-else:
Could you please explain the use case which requires this change?

or-else:
Yes, please remove .gitignore or move in to py_grpc

or-else:
Thanks!

PR #886, opened 2023-09-06
Adding address and diagnosys types

hadarhd:


or-else:
If you feel that something like this is necessary:

1. Explain what you are doing and why
2. 2. Do not make it specialized for your use case. It has to work for other people.

Issue #904, opened 2024-04-03
Drafty should correctly parse after fix for https://github.com/tinode/tinode-js/issues/74 


dilshans2k:
### Subject of the issue
The issue https://github.com/tinode/tinode-js/issues/74 suggests that each character of string must be treated as grapheme.

#### Server-side
- [x] Your own setup:
  * platform: Linux
  * version of Tinode server: 0.22
  * database backend: mongo
  * standalone

#### Client-side
- [x] TinodeWeb/tinodejs: javascript client
  * Browser make and version.
  * IMPORTANT! Use `index-dev.html` to reproduce the problem, not `index.html`.

### Steps to reproduce
On server, ensure that push notification adapter is turned on(fcm or tnpg).

Two ways to reproduce:
- Send a text message from iOS such that it contains emoji(multi codepoint) and some formatting. 
  For eg: `Hi 🏴󠁧󠁢󠁳󠁣󠁴󠁿 https://google.com`
- Apply [this](https://github.com/tinode/tinode-js/commit/ad52c9bea6a3d5f156b304a5f430f8a076e6f21f.patch) commit patch in tinode-js and send a similar message like above.

### Expected behaviour
Server should send push notification.

### Actual behaviour
Server doesn't send push notification and throws error, `fcm push: could not parse payload: invalid format`

PR #905, opened 2024-04-03
fix: #904 drafty should correctly parse message content for push notification

dilshans2k:
- instead of []rune, text is treated as []grapheme cluster
- added ut

### DIFF
--- a/server/drafty/drafty.go
+++ b/server/drafty/drafty.go
@@ -236,10 +236,6 @@ func toTree(drafty *document) (*node, error) {
 // Given a grapheme iterator, start and end pos, returns a new grapheme iterator
 // containing a slice of graphemes(start->end) from input interator.
 func sliceGraphemeClusters(g *uniseg.Graphemes, start, end int) *uniseg.Graphemes {
-	if g == nil {
-		return uniseg.NewGraphemes("")
-	}
-
 	g.Reset()
 
 	output := ""
@@ -270,9 +266,6 @@ func sliceGraphemeClusters(g *uniseg.Graphemes, start, end int) *uniseg.Grapheme
 
 // Given a grapheme iterator, returns the original string from which it was created from.
 func graphemeToString(g *uniseg.Graphemes) string {
-	if g == nil {
-		return ""
-	}
 	g.Reset()
 
 	output := ""
@@ -286,9 +279,6 @@ func graphemeToString(g *uniseg.Graphemes) string {
 
 // Returns the number of grapheme cluster found in iterator.
 func getGraphemeLength(g *uniseg.Graphemes) int {
-	if g == nil {
-		return 0
-	}
 	g.Reset()
 
 	output := 0
--- a/server/drafty/drafty.go
+++ b/server/drafty/drafty.go
@@ -233,8 +233,8 @@ func toTree(drafty *document) (*node, error) {
 	return &node{children: children}, nil
 }
 
-// given a grapheme iterator, start and end pos, returns a new grapheme iterator
-// containing a slice of graphemes(start->end) from input interator
+// Given a grapheme iterator, start and end pos, returns a new grapheme iterator
+// containing a slice of graphemes(start->end) from input interator.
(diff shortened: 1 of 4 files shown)


server/drafty/drafty.go
```diff
@@ -231,8 +233,102 @@ func toTree(drafty *document) (*node, error) {
 	return &node{children: children}, nil
 }
 
+// given a grapheme iterator, start and end pos, returns a new grapheme iterator
+// containing a slice of graphemes(start->end) from input interator
+func sliceGraphemeClusters(g *uniseg.Graphemes, start, end int) *uniseg.Graphemes {
+	if g == nil {
```
or-else, line 239: Can it be `nil`? How?

server/drafty/drafty.go
```diff
@@ -231,8 +233,102 @@ func toTree(drafty *document) (*node, error) {
...
+	}
+
+	g.Reset()
+
+	return uniseg.NewGraphemes(output)
+}
+
+// given a grapheme iterator, returns the original string from which it was created from
+func graphemeToString(g *uniseg.Graphemes) string {
+	if g == nil {
```
or-else, line 276: How can it be `nil`?

server/drafty/drafty.go
```diff
@@ -231,8 +233,102 @@ func toTree(drafty *document) (*node, error) {
...
+	for g.Next() {
+		output += g.Str()
+	}
+
+	return output
+}
+
+// returns the number of grapheme cluster found in iterator
+func getGraphemeLength(g *uniseg.Graphemes) int {
+	if g == nil {
```
or-else, line 292: How can it be `nil`?

server/drafty/drafty.go
dilshans2k: Yes, you are right, current code scenarios it is never assigned to nil, but since it is a pointer and somehow got assigned to nil, `g.Next()` would panic.

What I can do is return error if g is `nil`

server/drafty/drafty.go
```diff
@@ -231,8 +233,102 @@ func toTree(drafty *document) (*node, error) {
...
+			i++
+		} else if i == start {
+			// starting pos found, start collecting string
+			output = output + g.Str()
+			j = i + 1
+		}
+
+	}
+
+	g.Reset()
```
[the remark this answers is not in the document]
dilshans2k: Ok. I would call Reset before.

server/drafty/drafty.go
dilshans2k: Yes, you are right, current code scenarios it is never assigned to nil, but since it is a pointer and somehow got assigned to nil, g.Next() would panic.

What I can do is return error if g is nil

server/drafty/drafty.go
dilshans2k: Yes, you are right, current code scenarios it is never assigned to nil, but since it is a pointer and somehow got assigned to nil, g.Next() would panic.

What I can do is return error if g is nil

server/drafty/drafty_test.go
```diff
@@ -164,7 +182,7 @@ func TestPreview(t *testing.T) {
 		}
 		res, err := Preview(val, 15)
 		if err == nil {
-			t.Errorf("invalid input %d did not cause an error '%s'", testsToFail[i], res)
+			t.Errorf("invalid input %d did not cause an error '%s'", i, res)
```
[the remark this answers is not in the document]
dilshans2k: It is incorrect to put `testsToFail[i]` here.

Eg: Take the invalidInputs[5] and correct it such that line:185 would throw error. Because value of `i` at this stage is 5, and testsToFail length is only 4, so it would `panic: runtime error: index out of range [5] with length 4`

server/drafty/drafty.go
or-else: Let it panic.

or-else:
Any update on this?  Do you want to finish this PR?

or-else:
Is this PR abandoned?

dilshans2k:
I was just busy with something else. I will complete it soon.

or-else:
Thanks!

PR #906, opened 2024-04-05
CORS headers in upload file API for all types of requests

markeloff86:
There is a problem. If you host the backend and frontend on different domains and try to upload a file using the /v0/file/u/ API, you get the error: “Access to XMLHttpRequest at ‘https://backend/v0/file/u/’ from origin ‘https://frontend’ has been blocked by CORS policy: No ‘Access-Control-Allow-Origin’ header is present on the requested resource.”
This is happening because the Tinode chat server adds Access-Control-Allow-* headers only for OPTIONS requests, but not for POST. This scheme could have worked only if the POST request were a simple request, but in the POST request, we add custom headers such as X-Tinode-Apikey, X-Tinode-Auth, and it becomes not simple :)
So if our POST request isn’t a simple request, we have to add Access-Control-Allow-* headers also in OPTIONS and POST responses.

Examples in the attachments.
Screenshot 2024-04-05 at 22 21 28
Screenshot 2024-04-05 at 22 22 33

### DIFF
--- a/server/media/media.go
+++ b/server/media/media.go
@@ -98,11 +98,6 @@ func matchCORSMethod(allowMethods []string, method string) bool {
 
 // CORSHandler is the default preflight OPTIONS processor for use by media handlers.
 func CORSHandler(req *http.Request, allowedOrigins []string, serve bool) (http.Header, int) {
-	if req.Method != http.MethodOptions {
-		// Not an OPTIONS request. No special handling for all other requests.
-		return nil, 0
-	}
-
 	var allowMethods []string
 	if serve {
 		allowMethods = []string{http.MethodGet, http.MethodHead, http.MethodOptions}
--- a/server/media/s3/s3.go
+++ b/server/media/s3/s3.go
@@ -151,12 +151,8 @@ func (ah *awshandler) Init(jsconf string) error {
 	return err
 }
 
-// Headers redirects GET, HEAD requests to the AWS server.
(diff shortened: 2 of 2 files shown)


markeloff86:
Status code 200, but request was marked as error

markeloff86:
Status code 200, but request was marked as error
Screenshot 2024-04-05 at 22 27 10

or-else:
I agree that your use case (different domains for frontend and backend) is currently not working correctly, but this PR is at least incomplete. I believe your change 
https://github.com/tinode/chat/pull/906/files#diff-ce653c5ff815856431505be2421678e320d94dc64419911f9aaaaac2d6180c81L156-L159
will cause existing use cases to fail. For example `web.tinode.co` is setup to accept `POST` and redirect `GET`. 

Your change must be backwards-compatible. For instance, you can add configuration options to handle `GET` and `POST` differently, if necessary.

Your change https://github.com/tinode/chat/pull/906/files#diff-3854734b46c07689d296b362547fd828c11deaf89f709ac6a5abca7e483437e2L101-L105 is also not great because it will add a bunch of unnecessary headers to GET, POST, HEAD, etc requests, like `Access-Control-Allow-Methods`.

or-else:
I'm going to close this PR because it looks unsalvageable. Please take a look at https://github.com/tinode/chat/pull/907. It should solve your problem.

PR #907, opened 2024-04-09
Add CORS HTTP headers to regular (non-OPTIONS) requests

or-else:
Ref https://github.com/tinode/chat/pull/906

### DIFF
--- a/server/media/media.go
+++ b/server/media/media.go
@@ -109,6 +109,8 @@ func CORSHandler(req *http.Request, allowedOrigins []string, serve bool) (http.H
 	allowedOrigin := matchCORSOrigin(allowedOrigins, origin)
 	requestMethod := req.Header.Get("Access-Control-Request-Method")
 	if req.Method == http.MethodOptions && requestMethod != "" {
+		// Preflight request.
+
 		if allowedOrigin == "" {
 			return headers, http.StatusNoContent
 		}
--- a/server/media/media.go
+++ b/server/media/media.go
@@ -80,6 +80,7 @@ func matchCORSOrigin(allowed []string, origin string) string {
 	return ""
 }
 
+// allowMethods must be in UPPERCASE.
 func matchCORSMethod(allowMethods []string, method string) bool {
 	if method == "" {
 		// Request has no Method header.
@@ -88,7 +89,7 @@ func matchCORSMethod(allowMethods []string, method string) bool {
 
 	method = strings.ToUpper(method)
 	for _, mm := range allowMethods {
-		if strings.ToUpper(mm) == method {
+		if mm == method {
 			return true
 		}
 	}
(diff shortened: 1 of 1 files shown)


or-else:
@markeloff86 please check if it solves your problem.

markeloff86:
So now we are adding Access-Control-Allow-Origin for all method?
I think It might work

server/media/media.go
```diff
@@ -88,49 +89,59 @@ func matchCORSMethod(allowMethods []string, method string) bool {
...
+			allowMethods = []string{http.MethodPost, http.MethodPut, http.MethodHead, http.MethodOptions}
+		}
+
+		if !matchCORSMethod(allowMethods, requestMethod) {
+			// CORS policy does not allow this method.
+			return headers, http.StatusNoContent
+		}
+
+		headers["Access-Control-Allow-Headers"] = []string{"*"}
+		headers["Access-Control-Allow-Credentials"] = []string{"true"}
```
markeloff86, line 131: array?
The Access-Control-Allow-Credentials header expects a single string value, not a slice of strings

or-else:
> But I'm not sure about Access-Control-Allow-Credentials Should we add it too for all methods? Because we use auth headers https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Credentials

The way I read the spec it's not needed if request is preflighted and reflight includes `Access-Control-Allow-Credentials`:

* If the request is preflighted [... and] the server's response to the preflight request sets the Access-Control-Allow-Credentials header to true, then the real request will include credentials...

or-else:
Could you please just test it without your setup?

markeloff86:
Your new code doesn't add headers for POST and PUT
So I've made some changes and tested it
This works for me
https://github.com/tinode/chat/pull/909

or-else:
> it crashes here

How does it crash? Do you have a stack trace?

or-else:
> Your new code doesn't add headers for POST and PUT, because it crashes here
> 
> https://github.com/tinode/chat/blob/99c633297a805123f74642334f87d4a5b8ab7967/server/media/s3/s3.go#L164
> 
> 
> So I've made some changes and tested it
> This works for me
> #909

I don't see how it could possibly crash. I can see how it could return an error. Could you please elaborate.

markeloff86:
I'm not sure that's exacly problem in this line of code, but without condition which I've added for POST and PUT I always get 404 status code without Access-Control-Allow-Origin header.
So I presume 404 status code return this line of code https://github.com/tinode/chat/blob/99c633297a805123f74642334f87d4a5b8ab7967/server/media/s3/s3.go#L166

it returns because for POST and PUT we don't pass id. That's why this code shouldn't have been executed for POST and PUT.

or-else:
OK, that's what I thought. It's not a crash, but an error.

PR #915, opened 2024-08-10
Fix botUID in chatbot

THUzxj:
Hi,

Thanks for your great project. I am working on deploying Tinode and developing chatbots. The `botUID` in the current chatbot example does not work because it is always `None`. So this PR fixes it.

By the way, I can help build a unified Python API for both chatbot and tn-cli, based on the code of tn-cli. What is your opinion on this? Thank you!

### DIFF
--- a/chatbot/python/chatbot.py
+++ b/chatbot/python/chatbot.py
@@ -301,6 +301,7 @@ def read_auth_cookie(cookie_file_name):
     return schema, secret
 
 def on_login(cookie_file_name, params):
+    global botUID
     client_post(subscribe('me'))
 
     """Save authentication token to file"""
@@ -308,7 +309,7 @@ def on_login(cookie_file_name, params):
         return
 
     if 'user' in params:
-        botUID = params['user'].decode("ascii")
+        botUID = params['user'].decode("ascii")[1:-1]
 
     # Protobuf map 'params' is not a python object or dictionary. Convert it.
     nice = {'schema': 'token'}
--- a/chatbot/python/.gitignore
+++ b/chatbot/python/.gitignore
@@ -1,0 +1,1 @@
+.tn-cookie


chatbot/python/chatbot.py
```diff
@@ -256,7 +256,7 @@ def client_message_loop(stream):
                 exec_future(msg.ctrl.id, msg.ctrl.code, msg.ctrl.text, msg.ctrl.params)
 
             elif msg.HasField("data"):
-                # log("message from:", msg.data.from_user_id)
+                # log("message from:", msg.data.from_user_id, "botUID:", botUID)
```
or-else, line 259: I don't think this change is needed.

chatbot/python/chatbot.py
```diff
@@ -309,6 +310,8 @@ def on_login(cookie_file_name, params):
 
     if 'user' in params:
         botUID = params['user'].decode("ascii")
+        if botUID[0] == '"':
```
or-else, line 313: Why is this needed?

or-else:
Thank you for your PR. Please use `devel` branch.

chatbot/python/chatbot.py
THUzxj: Yes, it is just for debugging.

chatbot/python/chatbot.py
THUzxj: `params['user'].decode("ascii")` returns the string containing quotation marks in the front and end, like '"usrxxxxx"'. I am not sure it will always happen in any environments.

chatbot/python/chatbot.py
or-else: Yes, I understand, but you don't need to change an existing line. Or you can remove it alltogether.

chatbot/python/chatbot.py
or-else: It's always quoted because values `params` dictionary are bytes. So, just always unquote it:

```python
# Values params are bytes, so a string is always quoted. Unquote it.
botUID = params['user'].decode("ascii")[1:-1]
```

chatbot/python/chatbot.py
THUzxj: Removed the change.

chatbot/python/chatbot.py
THUzxj: Thanks, got it and fixed.

or-else (approved):

Issue #902, opened 2024-03-24
Cannot initialize postgresql database


kovmir:
I pre-created `tinode` database, but it fails saying the database already exists:

```
$ ./init-db -data=./data.json
2024/03/24 07:17:44 Database adapter: 'postgres'; version: 113
2024/03/24 07:17:44 Database not found. Creating.
2024/03/24 07:17:45 Failure: ERROR: database "tinode" already exists (SQLSTATE 42P04)
```

I dropped the database, but it still complains:

```
./init-db -data=./data.json
2024/03/24 07:19:53 Database adapter: 'postgres'; version: 113
2024/03/24 07:19:53 Failed to init DB adapter: failed to connect to `host=localhost user=tinode database=`: server error (FATAL: database "tinode" does not exist (SQLSTATE 3D000))
```

My `tinode.conf` snippet:

```
"postgres": {
                                // PostgreSQL connection settings.
                                "User": "tinode",
                                "Passwd": "ufuB1AdBSp-NtHuv",
                                "Host": "localhost",
                                "Port": "5432",
                                "DBName": "tinode",
                                
                                // DSN: alternative way of specifying database configuration, passed unchanged
                                // to the driver. See https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNSTRING
                                // "dsn": "postgresql://postgres:postgres@localhost:5432/tinode?sslmode=disable&connect_timeout=10",
```

Debian `bookworm`, tinode version `0.22.12`, PostgreSQL `16`.

or-else:
Unfortunately PgSQL is somewhat brain damaged: the database connection must specify an existing database, and if no database is provided, it defaults to the same database name as the connecting user name. That's why we provide sample connection as 
```
"User": "postgres",
"Passwd": "postgres",
``` 
The user `postgres` matches an existing database `postgres`.

If you do not want to use user `postgres`, then create another user with any other name than `tinode`. Say, create user `tinodeadmin`, create an empty database `tinodeadin` and everything will work.

kovmir:
>... create user `tinodeadmin`, create an empty database `tinodeadmin`

This produces exactly the same errors.

or-else:
Are you sure you did it right? Show it.

kovmir:
I finally got it, `User` and `DBName` within the config file must not match:

```
$ ./init-db
2024/03/24 17:59:36 Database adapter: 'postgres'; version: 113
2024/03/24 17:59:36 Database not found. Creating.
[...]

PR #916, opened 2024-08-14
Update README.md

lizhenwei:
postgres need create dabase befor run tinode service,and tinode service need set environment change host "localhost" to "postgres",postgres is container name

### DIFF
--- a/docker/README.md
+++ b/docker/README.md
@@ -51,7 +51,13 @@ All images are available at https://hub.docker.com/r/tinode/
 
 	2. **PostgreSQL**:
 	```
-	$ docker run -p 6060:6060 -d --name tinode-srv --network tinode-net tinode/tinode-postgres:latest
+ 	$ docker exec -it -u postgres postgres psql
+ 	psql (13.16 (Debian 13.16-1.pgdg120+1))
+	Type "help" for help.
+	
+	postgres=# create database tinode;
+ 	exit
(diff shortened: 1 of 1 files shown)


or-else:
The missing database is already handled here:
https://github.com/tinode/chat/blob/a97c49477fae478062238049e6cde2da2fe9c114/server/db/postgres/adapter.go#L148

If configured DB is missing, then adapter connects to 'default' database which is the same as the user name. We use `postgres` as username https://github.com/tinode/chat/blob/a97c49477fae478062238049e6cde2da2fe9c114/server/tinode.conf#L237 and database `postgres` always exists, consequently connection succeeds.

This string `--env POSTGRES_DSN=postgresql://postgres:postgres@postgres:5432/tinode` is the default setting. You basically just repeat what's already the default.

What you are suggesting is not harmful, but seems redundant. Or at least you need to explain the relationship between `tinode` in this line `postgres=# create database tinode;` and `tinode` in this line --env POSTGRES_DSN=postgresql://postgres:postgres@postgres:5432/tinode`, that it must be exactly the same database name.

See conversation here: https://github.com/tinode/chat/issues/902

or-else:
Or just give an explanation why the database needs to be created.

or-else:
Any update on this?

or-else:
I added an explanation to FAQ:
https://github.com/tinode/chat/blob/devel/docs/faq.md#q-how-to-fix-postgresql-initialization-failing-with-missing-database-error

I think it's enough.

or-else:
No update in while, no answers to questions. Closing.

PR #919, opened 2024-08-22
Update INSTALL.md

aheiabai:
in Running a Standalone Server section, there is no 'init-db', 'tinode' in $GOPATH/bin but tinode-db, server.


or-else:
Thanks

Issue #923, opened 2024-10-09
MariaDB support?


robertsilen:
Is MariaDB supported? Any reason not to mention MariaDB in the README? I'd be happy to submit a PR.

or-else:
MariaDB is supported to the extent that it's SQL and binary compatible with MySQL. I don't mind mentioning MariaDB compatibility in README but I don't want to mention it everywhere where MySQL is mentioned.

or-else:
I added a mention in https://github.com/tinode/chat/commit/777c2ab914609805e04c80c2d96438a773ee3761
Closing.

PR #924, opened 2024-10-10
add MariaDB to README.md

robertsilen:
Suggesting to add mention of MariaDB to README as discussed in https://github.com/tinode/chat/issues/923


or-else:
I would rather not mention it this way because it creates an impression that it's fully supported just like other databases. We do not support it explicitly and do not test on it. Given some differences in the way MySQL and MariaDB store JSON columns it's possible that they will become incompatible in the future. I would rather mention it as 

`MySQL (and MariaDB as long as it's SQL and binary compatible)`

robertsilen:
How about adding CI testing for both MariaDB and MySQL to verify support for both? Can you provide any guidance on CI for your project?

or-else:
It would be great. 
We have no CI consequently no guidance. And automated tests only for MongoDB.

or-else:
Any update on this?

PR #938, opened 2025-02-16
FIX-898: initial commit and test fix

AnthonyKot:
The initial attempt to display correctly number of unread messages and handle deleted message edge case.


or-else:
Sorry, but I don't think it's an acceptable fix - it's too expensive. You are going to read potentially multiple large messages from the DB for each `read` receipt. That alone would likely generate DB load comparable to everything else combined, particularly if there is a lot of large group topics.

AnthonyKot:
I agree. I'll treat this as a draft solution. My plan is to verify it locally and then work on adding a new query to MongoDB to efficiently retrieve the count of unread messages.

PR #951, opened 2025-06-15
chore/add ssl mode for postgres adapter

ercsnmrs:
- server: updated the tinode conf
- server/db/postgres: put default value


or-else:
Thanks!

PR #953, opened 2025-06-19
chore/add new index for sql optimizations

ercsnmrs:
- server/db/mysql: implemented version 115
- server/db/postgres: implemented version 115
- optimized the join for state filters and seqid
- optimized the search for relevant subscription

while load testing the postgres adpater, I found out that the CPU Usage is reaching ~100%
Screenshot 2025-06-16 at 20 28 26
we optimized this by finding doing an RCA and traced it  back to a query that reaches 2.5 second which is the function  **UserUnreadCount** 
After implementing the indexes, we ran the load test again and found significant improvements. 
Screenshot 2025-06-18 at 13 10 13

Test config: : 1 TPS | 30 min | 5000 customers and 5000 drivers as test data

Message flow: 
- Customer will login
- Cusomer will create a groupchat and add driver1 and driver2 to the chat
- Customer1 will publish first message
- Driver1 will login and publish second message
- Driver2 will login and publish third message

EXPLAIN QUERY RESULTS
Before
`GroupAggregate  (cost=2751.50..2751.53 rows=1 width=16)`
After
`GroupAggregate  (cost=0.70..83.51 rows=1 width=16)`

### DIFF
--- a/server/db/mysql/adapter.go
+++ b/server/db/mysql/adapter.go
@@ -45,7 +45,7 @@ const (
 	defaultDSN      = "root:@tcp(localhost:3306)/tinode?parseTime=true"
 	defaultDatabase = "tinode"
 
-	adpVersion = 114
+	adpVersion = 115
 
 	adapterName = "mysql"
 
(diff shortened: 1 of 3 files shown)


or-else:
Thanks! 

Could you please sign out the CLA: https://github.com/tinode/chat/blob/master/docs/CLA.md

ercsnmrs:
Already did the day before. Did one today now. Thanks!

PR #955, opened 2025-06-27
fix/postgres and topic function

ercsnmrs:
- server/topic: the service panics when try to use the search function on the webapp (try on xena)
- server/db/postgres: the findOne function does not return anything

Screenshot 2025-06-27 at 09 26 29
Screenshot 2025-06-27 at 09 26 20


or-else:
Thanks!

PR #957, opened 2025-07-10
Fixed the error message when using port 465 in SendEmail: SMTP error …

Dihubopen:
…xx@xx.com EOF

### DIFF
--- a/server/validate/email/validate.go
+++ b/server/validate/email/validate.go
@@ -394,7 +394,18 @@ func (v *validator) TempAuthScheme() (string, error) {
 
 // SendMail replacement
 func (v *validator) sendMail(rcpt []string, msg []byte) error {
-	client, err := smtp.Dial(v.SMTPAddr + ":" + v.SMTPPort)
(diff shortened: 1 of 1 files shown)


server/validate/email/validate.go
```diff
@@ -394,7 +394,18 @@ func (v *validator) TempAuthScheme() (string, error) {
...
 // SendMail replacement
 func (v *validator) sendMail(rcpt []string, msg []byte) error {
-	client, err := smtp.Dial(v.SMTPAddr + ":" + v.SMTPPort)
+	addr := v.SMTPAddr + ":" + v.SMTPPort
+	tlsConfig := &tls.Config{
+		InsecureSkipVerify: v.TLSInsecureSkipVerify,
+		ServerName:         v.SMTPAddr,
+	}
+
+	conn, err := tls.Dial("tcp", addr, tlsConfig)
```
or-else, line 403: This call will fail if SMTP is still using insecure connection (port 25).

PR #969, opened 2025-08-31
feature/integrate gcs kms bucket for storing images

ercsnmrs:
Overview:
- KMS-protected buckets gives full control over encryption keys (vs Google-managed keys), granular permissions: You can restrict access to specific keys per team or service, and rotation & auditability: You can rotate keys periodically and audit access using Cloud Logging.


ercsnmrs:
implemented the same approach on an enterprise solution 
- I will test it out first by removing the limitation on maxInbandAttachmentSize on the Webapp
- I will post the results here.

or-else:
Cool, thanks.

> removing the limitation on maxInbandAttachmentSize on the Webapp

I'm not sure what you mean by this. Do you want to set `maxInbandAttachmentSize` to zero?

ercsnmrs:
@or-else i am trying to test the upper limit of the integration (e.g rps on which the integration to kms buckets break).

One use case i see here is that we can offload all images and attachment to the cloud instead of saving to db. I see it as a path for optimization here. 

- [ ] To do this, we need to send everything to gcs by lowering the maxInBandAttachmensize or disabling (i will do this locally)

or-else:
OK, makes sense.

or-else:
Should I review this PR too?

PR #988, opened 2026-02-26
perf(push/tnpg): reduce allocations in postMessage with sync.Pool

### LINKED ISSUE #987: Performance: reuse bytes.Buffer (sync.Pool) in TNPG postMessage to reduce allocations
- Context: In server/push/tnpg/push_tnpg.go , postMessage currently creates a new bytes.Buffer (and gzip writer) on every call. Under high push traffic this can increase heap allocations and GC pressure.
- Suggestion: Introduce a sync.Pool for *bytes.Buffer (optionally also for *gzip.Writer ) and reuse it across calls.
- Important detail: Always call buf.Reset() after Get() (and reset gzip writer state if pooled) to avoid reusing stale data.
- Expected benefit: Fewer heap allocations, reduced GC overhead, lower latency during bursty push batches.
- Notes: sync.Pool is best-effort and may be cleared by GC, which is fine here because it’s an optimization, not correctness-critical.

Antoanyabo:
- In server/push/tnpg/push_tnpg.go (postMessage), reuse *bytes.Buffer and *gzip.Writer via sync.Pool to reduce per-request allocations and GC pressure.
- Buffers are reset on reuse, and very large buffers are not kept in the pool to avoid memory bloat.
- Fixes #987

### DIFF
--- a/server/push/tnpg/push_tnpg.go
+++ b/server/push/tnpg/push_tnpg.go
@@ -36,14 +36,14 @@ var handler Handler
 const maxPooledPostBodyCap = 1 << 16
 
 var postBodyPool = sync.Pool{
-	New: func() interface{} {
+	New: func() any {
 		return new(bytes.Buffer)
 	},
 }
 
 var gzipWriterPool = sync.Pool{
-	New: func() interface{} {
-		return gzip.NewWriter(io.Discard)
+	New: func() any {
+		return gzip.NewWriter(nil)
 	},
 }
 
@@ -169,7 +169,7 @@ func (Handler) Init(jsonconf json.RawMessage) (bool, error) {
 	return true, nil
 }
 
-func postMessage(endpoint string, body interface{}, config *configType) (*batchResponse, error) {
+func postMessage(endpoint string, body any, config *configType) (*batchResponse, error) {
 	buf := postBodyPool.Get().(*bytes.Buffer)
 	defer func() {
 		buf.Reset()
@@ -181,7 +181,7 @@ func postMessage(endpoint string, body interface{}, config *configType) (*batchR
 
 	gzw := gzipWriterPool.Get().(*gzip.Writer)
 	defer func() {
-		gzw.Reset(io.Discard)
+		gzw.Reset(nil)
 		gzipWriterPool.Put(gzw)
 	}()
 	gzw.Reset(buf)
--- a/server/push/tnpg/push_tnpg.go
+++ b/server/push/tnpg/push_tnpg.go
@@ -171,7 +171,6 @@ func (Handler) Init(jsonconf json.RawMessage) (bool, error) {
 
 func postMessage(endpoint string, body interface{}, config *configType) (*batchResponse, error) {
 	buf := postBodyPool.Get().(*bytes.Buffer)
-	buf.Reset()
 	defer func() {
 		buf.Reset()
 		if cap(buf.Bytes()) > maxPooledPostBodyCap {
@@ -181,12 +180,15 @@ func postMessage(endpoint string, body interface{}, config *configType) (*batchR
 	}()
 
 	gzw := gzipWriterPool.Get().(*gzip.Writer)
+	defer func() {
+		gzw.Reset(io.Discard)
+		gzipWriterPool.Put(gzw)
+	}()
 	gzw.Reset(buf)
 	err := json.NewEncoder(gzw).Encode(body)
 	if closeErr := gzw.Close(); err == nil {
 		err = closeErr
 	}
-	gzipWriterPool.Put(gzw)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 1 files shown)


or-else:
Please sign the CLA: https://github.com/tinode/chat/blob/master/CONTRIBUTING.md

Antoanyabo:
> Please sign the CLA: https://github.com/tinode/chat/blob/master/CONTRIBUTING.md

already done, sir !

server/push/tnpg/push_tnpg.go
```diff
@@ -154,11 +169,24 @@ func (Handler) Init(jsonconf json.RawMessage) (bool, error) {
...
+	defer func() {
+		buf.Reset()
+		if cap(buf.Bytes()) > maxPooledPostBodyCap {
+			return
+		}
+		postBodyPool.Put(buf)
+	}()
+
+	gzw := gzipWriterPool.Get().(*gzip.Writer)
+	gzw.Reset(buf)
```
or-else, line 187: The `gzw` saves the pointer to `buf`.

server/push/tnpg/push_tnpg.go
```diff
@@ -154,11 +169,24 @@ func (Handler) Init(jsonconf json.RawMessage) (bool, error) {
...
 	err := json.NewEncoder(gzw).Encode(body)
-	gzw.Close()
+	if closeErr := gzw.Close(); err == nil {
+		err = closeErr
+	}
+	gzipWriterPool.Put(gzw)
```
or-else: Here you return `gzw` to the pool with the reference to `buf`. Which means GC cannot touch `buf` even if it's no longer needed. Why not call `gzw.Reset(nil)` first?

server/push/tnpg/push_tnpg.go
```diff
@@ -154,11 +169,24 @@ func (Handler) Init(jsonconf json.RawMessage) (bool, error) {
 	return true, nil
 }
 
-func postMessage(endpoint string, body any, config *configType) (*batchResponse, error) {
-	buf := new(bytes.Buffer)
-	gzw := gzip.NewWriter(buf)
+func postMessage(endpoint string, body interface{}, config *configType) (*batchResponse, error) {
+	buf := postBodyPool.Get().(*bytes.Buffer)
+	buf.Reset()
```
or-else: You are resetting it twice, after use and then again before use. Is it really needed?

server/push/tnpg/push_tnpg.go
```diff
@@ -32,6 +33,20 @@ const (
...
+
+var postBodyPool = sync.Pool{
+	New: func() interface{} {
+		return new(bytes.Buffer)
+	},
+}
+
+var gzipWriterPool = sync.Pool{
+	New: func() interface{} {
+		return gzip.NewWriter(io.Discard)
```
or-else: `gzip.NewWriter(nil)` is perfectly fine too.

server/push/tnpg/push_tnpg.go
Antoanyabo: Good catch — buf.Reset() is currently called twice. It's redundant if we ensure buffers are reset before Put().
I'll remove the pre-use Reset() and keep the deferred cleanup Reset() before returning to the pool.

server/push/tnpg/push_tnpg.go
Antoanyabo: I used io.Discard as a safe placeholder; nil works too since we Reset before use. Happy to change to nil if you prefer

server/push/tnpg/push_tnpg.go
Antoanyabo: Agreed. I now reset the gzip.Writer before returning it to the pool to drop the reference to buf (avoid retention). I used Reset(io.Discard) as a safe placeholder; Reset(nil) would work too since we always Reset(buf) before use.

server/push/tnpg/push_tnpg.go
Antoanyabo: Agreed. I now reset the gzip.Writer before returning it to the pool to drop the reference to buf (avoid retention). I used Reset(io.Discard) as a safe placeholder; Reset(nil) would work too since we always Reset(buf) before use.

server/push/tnpg/push_tnpg.go
or-else: Yes, please do. I would rather have a panic than a quiet failure if the writer is used incorrectly.

or-else:
Please replace all `interface{}` with `any`.

server/push/tnpg/push_tnpg.go
Antoanyabo: > Yes, please do. I would rather have a panic than a quiet failure if the writer is used incorrectly.

done!

Antoanyabo:
> Please replace all `interface{}` with `any`.

done!

Issue #316, opened 2019-11-20
[Question]deploy on kubernetes


diaobrother:
On kubernetes, every node is the same. In this case, how to make communication between nodes?

or-else:
We con't use k8s exactly for this reason. We deploy three docker containers using docker-compose.

ysaakpr:
Does this mean tinode shall not be deployed on Kubernetes? I believe, the application should not restrict the deployment infrastructure. Due to increased adoption of Kubernetes, it would be better to publish the way to deploy on Kubernetes, Possible ways of cluster discovery
1. Can use service dns to discover all the IP addresses which serving at a point in time.
2. A discovery announcer agent can be used as a side can which frequent;y poll either service or pods for getting all the cluster nodes which are not joined the cluster and can add to all the cluster.
3. A state can be maintained separately using things like consul or etcd and can figureout cluster using the state from them.

or-else:
As I wrote earlier, deploying Tinode with k8s makes no sense to me because k8s expects stateless identical pods. K8s is a wrong tool for the job. But of course you are welcome to deploy it with k8s if you see a use case for yourself.

1. This is true but not relevant.
2. Cluster nodes must be configured ahead of time to discover each other. Tinode is not stateless like a web server. 
3. I don't think so.

Tonkonozhenko:
@or-else we are going to use tinode soon. And I'd like to deploy it in k8s. I'm going to use external DB for storage, s3 for assets. what other stateful stuff do you have that you say it's not possible to deploy it in k8s?

or-else:
If you are deploying a cluster then every node maintains a state

or-else:
Each node is different from every other cluster node. Like when you run a mysql server, you don't deploy multiple pods with the same database, right? Same here. 

If you are deploying a standalone server, then you can use k8s to deploy a single instance of it.

Tonkonozhenko:
@or-else I found a guide [here](https://github.com/tinode/chat/blob/master/INSTALL.md#running-a-cluster) that tells that we need names for each node.
My suggestion is to deploy it using k8s stateful sets with fixed names.

or-else:
Sure, please do it and share your experience.

Tonkonozhenko:
@or-else sure, I will. Does tinode have something stored in FS so I need put it to volume?

or-else:
If you use s3 for user-generated content, then the only thing it reads from the file system is static config (read-only).

ysaakpr:
[...]

PR #994, opened 2026-04-28
K8s based service discovery

kaustubhbabar5:
# Kubernetes Discovery PR Summary (#316)

## Summary

Adds Kubernetes-native cluster discovery for Tinode. Users can now run Tinode as
a `StatefulSet` and let pods discover peers automatically from the headless
Service `EndpointSlice`, instead of maintaining a static
`cluster_config.nodes` list.

## How To Use

Apply the provided manifests:

```bash
kubectl apply -f docker/k8s/
kubectl rollout status statefulset/tinode
```

Enable Kubernetes discovery in `tinode.conf`:

```json
"cluster_config": {
  "self": "",
  "discovery": {
    "mode": "kubernetes",
    "kubernetes": {
      "namespace": "",
      "service": "tinode-headless",
      "port_name": "cluster"
    }
  }
}
```

Scale the cluster with Kubernetes:

```bash
kubectl scale statefulset/tinode --replicas=5
```

## Notes

- Requires a `StatefulSet`; `Deployment` is not supported because Tinode derives
  stable worker IDs from pod ordinals.
- Requires the `POD_NAME` downward API environment variable.
- The headless Service must expose the cluster RPC port with the name `cluster`.
- Static `nodes` and `failover` config are ignored in Kubernetes discovery mode.
- Cluster size can be checked through `TotalClusterNodes` in stats.


kaustubhbabar5:
@or-else requesting your review for this

or-else:
I'll try to review this weekend.

or-else:
@kaustubhbabar5 while waiting for my review, please address copilot comments. 
Did you sign the CLA?
https://github.com/tinode/chat/blob/master/docs/CLA.md

Thanks!

Abeabenut (commented):
kaustubhbabar5:k8s_based_service_discovery

or-else:
I promised to review sooner, but I've been really busy for the last two weeks. I'll try to find time this week

docker/tinode/Dockerfile.k8s
```diff
@@ -0,0 +1,41 @@
+FROM golang:1.26 AS build
```
or-else, line 1: No attempt is made to support storage options other than postgres in this dockerfile.

A better option would be to build the binary with `-tag mysql,mongodb,rethinkdb,postgres` or (worse) rename it to `Dockerfile.pgsql.k8s`.

docker/tinode/entrypoint-k8s.sh
```diff
@@ -0,0 +1,158 @@
+#!/bin/bash
```
or-else, line 1: This looks like largely a copy-paste of `entrypoint.sh`. Consider importing `entrypoint.sh` and then adding logic specific to k8s.

server/db/mysql/adapter.go
```diff
@@ -195,6 +195,21 @@ func (a *adapter) IsOpen() bool {
 	return a.db != nil
 }
 
+func (a *adapter) getHealthContext() (context.Context, context.CancelFunc) {
```
or-else, line 198: Making it a separate method looks unnecessary. It's called from one location and it's tiny. Just inline in where it's called.

server/db/mysql/adapter.go
```diff
@@ -195,6 +195,21 @@ func (a *adapter) IsOpen() bool {
...
+	if a.sqlTimeout > 0 {
```
or-else, line 199: This looks like an inconsistency between mongo+rethink vs mysql+pgsql. I would rather have it behave identically in all adapters.

server/cluster.go
```diff
@@ -102,24 +113,94 @@ type ClusterNode struct {
 }
 
 func (n *ClusterNode) asyncRpcLoop() {
```
or-else, line 115: Please add comments explaining the logic.

server/cluster.go
```diff
@@ -102,24 +113,94 @@ type ClusterNode struct {
...
-		if req == nil {
-			// Stop
-			return
-		}
+	for {
+		select {
+		case <-n.done:
+			for {
+				select {
+				case req, ok := <-n.p2mSender:
```
or-else, line 171: This block is copy-pasted below. Please use a single block of code (make it a function call?)

server/cluster.go
```diff
@@ -307,16 +410,21 @@ func (n *ClusterNode) reconnect() {
...
 	}
 
-	if err := n.endpoint.Call(proc, req, resp); err != nil {
+	if err := endpoint.Call(proc, req, resp); err != nil {
 		logs.Warn.Println("cluster: call failed", n.name, err)
 
 		n.lock.Lock()
-		if n.connected {
-			n.endpoint.Close()
+		if n.connected && !n.retired {
```
or-else, line 426: Checking both is redundant. if `node.connected == true` then `n.retired == false`.

server/cluster.go
```diff
@@ -307,16 +410,21 @@ func (n *ClusterNode) reconnect() {
 }
 
 func (n *ClusterNode) call(proc string, req, resp any) error {
-	if !n.connected {
+	n.lock.Lock()
+	retired := n.retired
+	connected := n.connected
+	endpoint := n.endpoint
+	n.lock.Unlock()
+	if retired || !connected {
```
or-else, line 418: This check is redundant. You only need to check `connected`.

You also don't need to allocate vars. You can check `!n.connected` then return with `defer n.lock.Unlock()`.

server/cluster.go
```diff
@@ -307,16 +410,21 @@ func (n *ClusterNode) reconnect() {
...
-	if err := n.endpoint.Call(proc, req, resp); err != nil {
+	if err := endpoint.Call(proc, req, resp); err != nil {
 		logs.Warn.Println("cluster: call failed", n.name, err)
 
 		n.lock.Lock()
-		if n.connected {
-			n.endpoint.Close()
+		if n.connected && !n.retired {
+			n.closeEndpointLocked()
 			n.connected = false
```
or-else, line 428: It looks like every time you call `closeEndpointLocked` you immediately set `n.connected = false`, so you can move `n.connected = false` to `closeEndpointLocked`.

server/cluster.go
```diff
@@ -347,7 +455,9 @@ func (n *ClusterNode) callAsync(proc string, req, resp any, done chan *rpc.Call)
 		logs.Err.Panic("cluster: RPC done channel is unbuffered")
 	}
 
-	if !n.connected {
+	n.lock.Lock()
+	if n.retired || !n.connected {
```
or-else, line 459: Checking connected is enough.

server/cluster.go
```diff
@@ -360,6 +470,7 @@ func (n *ClusterNode) callAsync(proc string, req, resp any, done chan *rpc.Call)
 		}
 		return call
 	}
+	endpoint := n.endpoint
```
or-else, line 473: Why do you need to copy?

server/cluster.go
```diff
@@ -347,7 +455,9 @@ func (n *ClusterNode) callAsync(proc string, req, resp any, done chan *rpc.Call)
 		logs.Err.Panic("cluster: RPC done channel is unbuffered")
 	}
 
-	if !n.connected {
+	n.lock.Lock()
```
or-else, line 458: Consider `defer n.lock.Unlock()`

server/cluster.go
```diff
@@ -395,8 +506,15 @@ func (n *ClusterNode) proxyToMaster(msg *ClusterReq) error {
 
 // proxyToMaster forwards request from topic proxy to topic master.
 func (n *ClusterNode) proxyToMasterAsync(msg *ClusterReq) error {
+	n.lock.Lock()
+	if n.retired {
```
or-else, line 510: Why not check `connected` instead?

server/cluster.go
```diff
@@ -395,8 +506,15 @@ func (n *ClusterNode) proxyToMaster(msg *ClusterReq) error {
 
 // proxyToMaster forwards request from topic proxy to topic master.
 func (n *ClusterNode) proxyToMasterAsync(msg *ClusterReq) error {
+	n.lock.Lock()
```
or-else, line 509: Consider `defer n.lock.Unlock()`

server/cluster.go
```diff
@@ -781,14 +927,16 @@ func (c *Cluster) routeUserReq(req *UserCacheReq) error {
 
 // Given topic name, find appropriate cluster node to route message to.
 func (c *Cluster) nodeForTopic(topic string) *ClusterNode {
+	c.nodesLock.RLock()
```
or-else, line 930: Consider defer unlock.

server/cluster.go
```diff
@@ -801,7 +949,22 @@ func (c *Cluster) isRemoteTopic(topic string) bool {
...
 	}
-	return c.ring.Get(topic) != c.thisNodeName
+	_, remote := c.topicOwner(topic)
+	return remote
+}
+
+// topicOwner returns the cluster node that owns the topic and whether that
+// owner is remote, using a single ring snapshot.
+func (c *Cluster) topicOwner(topic string) (string, bool) {
+	if c == nil {
```
or-else, line 959: This looks redundant. The c.topicOwner is always called with c != nil checked first.

server/cluster.go
```diff
@@ -988,53 +1160,165 @@ func clusterInit(configString json.RawMessage, self *string) int {
...
+func resolvePodName(mode string, self *string, configThisName string) string {
+	// In kubernetes mode the node name is derived from the pod (not from the
+	// static config / CLI flag), and the listen address is bound to all
+	// interfaces since peers dial via the pod IP published in EndpointSlices.
+	if mode == "kubernetes" {
+		if *self != "" || configThisName != "" {
+			logs.Warn.Println("Cluster: ignoring cluster_self / self config in kubernetes mode; identity is derived from POD_NAME")
+		}
+
+		thisName := os.Getenv("POD_NAME")
```
or-else, line 1265: I would rather not introduce the direct reading from the env. It's not used anywhere else in the project. Please read the env in the entrypoint.sh, then pass it as a command line flag.

server/cluster.go
```diff
@@ -988,53 +1160,165 @@ func clusterInit(configString json.RawMessage, self *string) int {
 		logs.Warn.Println("Cluster config: field num_proxy_event_goroutines is deprecated.")
 	}
 
+	var discovery Discovery
```
or-else, line 1163: I think it's best to move the k8s & discovery logic to a separate file.

server/cluster.go
```diff
@@ -1101,35 +1504,135 @@ func (c *Cluster) shutdown() {
...
+// Kept as a helper for targeted runtime add/remove operations and unit tests.
+// The Kubernetes reconciler publishes full membership snapshots via
+// applySnapshot instead of mutating the peer set incrementally. Does nothing
+// and returns nil if a node with `name` already exists.
+func (c *Cluster) addClusterNode(name, addr string) *ClusterNode {
+	if name == c.thisNodeName {
+		return nil
+	}
+
+	c.nodesLock.Lock()
```
or-else, line 1552: defer unlock?

server/cluster.go
```diff
@@ -1182,18 +1687,44 @@ func (c *Cluster) gcProxySessions(activeNodes []string) {
...
-	n.msess = make(map[string]struct{})
-	n.lock.Unlock()
+	c.nodesLock.RUnlock()
+	if n == nil {
+		return
+	}
+	c.gcProxySessionsOnNode(n)
+}
+
+func (c *Cluster) gcProxySessionsOnNode(node *ClusterNode) {
```
or-else, line 1699: Please add comments

